PR #115, opened 2021-03-13
Add docker npm scripts, update docs

trxcllnt:
* Adds scripts to build/run the containers, and update the docs with the new commands:
  ```
  yarn docker:build:devel
  yarn docker:run:devel
  ```
  In addition to being simpler than invoking `docker-compose` directly, these commands will also `touch .env` beforehand, to ensure the `.env` file is created before docker-compose attempts to read it.

* Deletes the current `.env` file, which should hopefully go back to being gitignored locally.
* Adds the new configuration options to `docker-compose.yml` which should allow docker-compose v1.28.5+ to use the `"nvidia"` runtime without needing to set `"default-runtime": "nvidia"` in `/etc/docker/daemon.json`. Instructions on how to modify this file for older versions of docker-compose have been left in place.

Branched from https://github.com/rapidsai/node-rapids/pull/114, so we should merge it first.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,12 @@ x-common-service-settings: &common_service_settings
     XAUTHORITY: "${XAUTHORITY:-}"
     XDG_SESSION_TYPE: "${XDG_SESSION_TYPE:-}"
     NVIDIA_DRIVER_CAPABILITIES: "all"
-
+  deploy:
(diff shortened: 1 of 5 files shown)


ajschmidt8 (approved):
could use some more semi-colons, but otherwise LGTM :+1:

PR #116, opened 2021-03-13
Fix groupBy with multiple value columns

trxcllnt:
This PR fixes a small oversight in the groupBy aggregation construction.

`cudf::groupby` needs a list of `cudf::groupby::aggregation_request` for each values column, and each `aggregation_request` has a list of `unique_ptr<cudf::groupby::aggregation>`.

The previous logic was constructing just one `unique_ptr<cudf::groupby::aggregation>` and moving it into each column's `aggregation_request`.

This doesn't work because `std::move()` is a destructive operation -- after the `agg` argument was moved into the first column's `aggregation_request` it was left in an invalid state, causing all the rest of the aggregation requests to receive an invalid `unique_ptr<aggregation>`.

~Branched from https://github.com/rapidsai/node-rapids/pull/114 and https://github.com/rapidsai/node-rapids/pull/115, which we need to merge first.~


ajschmidt8 (approved):

AjayThorve (approved):
looks good

PR #118, opened 2021-03-15
Expand cuda/RMM tests and docs

trxcllnt:
* Expands and cleans up docs in the CUDA and RMM modules
* Adds more tests and functionality to the `MemoryView` classes

Branched from https://github.com/rapidsai/node-rapids/pull/117


modules/cuda/src/util.ts
```diff
@@ -59,13 +64,38 @@ export const isArrayBufferView = ArrayBuffer.isView;
...
+ *
+ * clampRange(5)        // [0, 5]
+ * clampRange(5, 0, -1) // [0, 4]
+ * clampRange(5, -1)    // [4, 5]
+ * clampRange(5, -1, 0) // [4, 4]
+ *
+ * const ary = Array.from({length: 5}, (_, i) => i);
+ * // [0, 1, 2, 3, 4]
+ * assert(ary.slice() == ary.slice(...clampRange(ary.length)))
+ * // [0, 1, 2, 3]
```
bryevdv, line 82: why wouldn't this be `[0,1,2,3,4]` if `clampRange(5)` is `[0, 5]` (above)

bryevdv (approved):

bryevdv:
LGTM just the one doc comment

modules/cuda/src/util.ts
trxcllnt: Oh this is my bad. The comments are above the line that produces them. I'll change that so there's no confusion.

trxcllnt: Done :+1:

PR #119, opened 2021-03-15
Add GH action for docs

ajschmidt8:
This PR adds a GitHub Action for automatically updating the `gh-pages` branch with the latest docs.

It works by checking out both the `main` and `gh-pages` branches to separate directories, doing a doc build in the `main` directory, copying the results to the `gh-pages` directory, then finally committing/pushing the changes to the `gh-pages` branch.

I added the `paths` key to try to eliminate excessive builds from irrelevant file changes, but this can be tweaked/removed if necessary (file matching patterns are [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-file-paths))

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,47 @@
+name: docs
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**/typedoc.js'
+      - '**/tsconfig.json'
+      - '**/package.json'
+      - 'modules/**'
+      - '.github/workflows/**'
+      - '!**/tests/**'
+
+jobs:
+  update:
+    name: Update Docs
+    runs-on: ubuntu-20.04
+    container: node:14
+    steps:
+      - name: Checkout main
+        uses: actions/checkout@v2
+        with:
+          path: main
+      - name: Checkout gh-pages
+        uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+          path: gh-pages
+      - name: Build Docs
+        working-directory: main
+        run: yarn && yarn tsc:build && yarn doc
+      - name: Copy Docs to gh-pages
+        run: |
+          rm -rf gh-pages/docs
+          cp -r main/doc gh-pages/docs
+      - name: Commit new docs
+        working-directory: gh-pages
+        run: |
+          COMMIT_SHA=$(echo "${COMMIT_SHA}" | cut -c 1-7)
+          git config user.name "GitHub Actions"
+          git config user.email "actions@github.com"
+          git add --all
+          git commit --allow-empty -m "update docs from ${COMMIT_SHA}"
+          git push gh-pages
+        env:
+          COMMIT_SHA: ${{ github.sha }}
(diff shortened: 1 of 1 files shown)


AjayThorve:
@ajschmidt8 thanks for this!! Just had a quick couple of question.

Will this action be executed locally? What if we create a PR, and then it merges with `main`, what triggers this action?

ajschmidt8:
> @ajschmidt8 thanks for this!! Just had a quick couple of question.
> 
> Will this action be executed locally? What if we create a PR, and then it merges with `main`, what triggers this action?

@AjayThorve, no problem! The file in this PR is a configuration file for a [GitHub Action](https://docs.github.com/en/actions), which is GitHub's CI/CD platform. It will be executed by GitHub on their own hosted hardware ([hardware docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners)). [These lines](https://github.com/rapidsai/node-rapids/blob/5b88f0852ab1967942c144bc6b6e0c0b028f7210/.github/workflows/docs.yaml#L3-L13) from the file determine when the automation will be run, which in this case is anytime a `push` is made to the `main` branch (i.e. a PR is merged). Additionally, it will only execute when that push includes changes to the files listed in the `paths` key (though we can update/change the values I added here if necessary).

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**/typedoc.js'
+      - '**/tsconfig.json'
+      - '**/package.json'
+      - 'modules/**'
```
AjayThorve, line 11: Since its just the JavaScript/TypeScript api documentation, can this be `modules/*.ts`?

thoughts? @trxcllnt @bryevdv @ajschmidt8

AjayThorve (commented):
> @AjayThorve, no problem! The file in this PR is a configuration file for a GitHub Action, which is GitHub's CI/CD platform. It will be executed by GitHub on their own hosted hardware (hardware docs). These lines from the file determine when the automation will be run, which in this case is anytime a push is made to the main branch (i.e. a PR is merged). Additionally, it will only execute when that push includes changes to the files listed in the paths key (though we can update/change the values I added here if necessary).

@ajschmidt8 sounds good,  thanks!

jolorunyomi (approved):

.github/workflows/docs.yaml
trxcllnt: Yeah a `modules/**/*.ts` should work!

trxcllnt (approved):

.github/workflows/docs.yaml
ajschmidt8: sounds good, i updated it to the following:

```yaml
paths:
      - '**/typedoc.js' # typedoc config changes
      - 'modules/**/src/**/*.ts' # src files, but not unit tests
      - '.github/workflows/**' # useful for ensuring that changes to the workflow file run successfully right after they're committed
```

PR #121, opened 2021-03-16
Add regex series functions

### LINKED ISSUE #120: Wrap libcudf regex functions
ref: https://docs.rapids.ai/api/libcudf/stable/md_regex.html

bryevdv:
issues: closes #120 

```node
> a = Arrow.Utf8Vector.from(["foo bar baz", "baz bar foo", "foo bar foo", "baz quux"])

> s = cudf.StringSeries.new(a)
StringSeries { _col: Column {} }
> [...s.containsRe("foo").toArrow()]
[ true, true, true, false ]
> [...s.countRe("foo").toArrow()]
[ 1, 1, 2, 0 ]
> [...s.matchesRe("foo").toArrow()]
[ true, false, true, false ]
```

Still needs tests and `findall_re` (or a separate issue)

### DIFF
--- a/modules/cudf/src/column/re.cpp
+++ b/modules/cudf/src/column/re.cpp
@@ -1,0 +1,72 @@
+// Copyright (c) 2021, NVIDIA CORPORATION.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <node_cudf/column.hpp>
+#include <node_cudf/utilities/napi_to_cpp.hpp>
+
+#include <node_rmm/memory_resource.hpp>
+
+#include <cudf/unary.hpp>
+
+#include <rmm/mr/device/per_device_resource.hpp>
+#include "cudf/strings/contains.hpp"
+
+namespace nv {
+
+ObjectUnwrap<Column> Column::contains_re(std::string const& pattern,
+                                         rmm::mr::device_memory_resource* mr) const {
+  try {
+    return Column::New(cudf::strings::contains_re(this->view(), pattern, mr));
+  } catch (cudf::logic_error const& err) { NAPI_THROW(Napi::Error::New(Env(), err.what())); }
+}
+
+ObjectUnwrap<Column> Column::count_re(std::string const& pattern,
+                                      rmm::mr::device_memory_resource* mr) const {
+  try {
+    return Column::New(cudf::strings::count_re(this->view(), pattern, mr));
+  } catch (cudf::logic_error const& err) { NAPI_THROW(Napi::Error::New(Env(), err.what())); }
+}
+
+ObjectUnwrap<Column> Column::matches_re(std::string const& pattern,
+                                        rmm::mr::device_memory_resource* mr) const {
+  try {
+    return Column::New(cudf::strings::matches_re(this->view(), pattern, mr));
+  } catch (cudf::logic_error const& err) { NAPI_THROW(Napi::Error::New(Env(), err.what())); }
+}
+
+Napi::Value Column::contains_re(Napi::CallbackInfo const& info) {
+  if (info.Length() < 1) {
+    NODE_CUDF_THROW("Column contains_re expects a pattern and optional MemoryResource", info.Env());
+  }
+  return contains_re(NapiToCPP{info[0]},
+                     NapiToCPP(info[1]).operator rmm::mr::device_memory_resource*());
+}
+
+Napi::Value Column::count_re(Napi::CallbackInfo const& info) {
+  if (info.Length() < 1) {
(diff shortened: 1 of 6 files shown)


modules/cudf/src/column/re.cpp
```diff
@@ -0,0 +1,72 @@
...
+  try {
+    return Column::New(cudf::strings::matches_re(this->view(), pattern, mr));
+  } catch (cudf::logic_error const& err) { NAPI_THROW(Napi::Error::New(Env(), err.what())); }
+}
+
+Napi::Value Column::contains_re(Napi::CallbackInfo const& info) {
+  if (info.Length() < 1) {
+    NODE_CUDF_THROW("Column contains_re expects a pattern and optional MemoryResource", info.Env());
+  }
+  return contains_re(NapiToCPP{info[0]},
```
trxcllnt, line 52: It's possible the JS engine coerces `RegExp` to strings when we do `val.ToString()` but we should double check. I could easily see someone doing `str.containsRe(/(foo|bar)/ig)` thinking it should work.

If it doesn't work, we can add a case to handle converting `RegExp` to strings in the `NapiToCPP` struct `std::string` conversion operator overload.

bryevdv: Well, currently it doesn't blow up, but it doesn't work either
```
> [...s.containsRe(/foo/ig).toArrow()]
[ false, false, false, false ]
```

bryevdv: But also neither does this
```
> [...s.containsRe("/foo/ig").toArrow()]
[ false, false, false, false ]
```
Unless we have some reliable way to automatically convert `/foo/ig` in to the specific set of "regex features" syntax here:

https://docs.rapids.ai/api/libcudf/stable/md_regex.html

we might be better of just raising an error that includes a link to those docs.

bryevdv: For `m` it seems as though `libcudf` always behaves with multi-line enabled, I am not sure there is any way to turn it *off*.

trxcllnt: Oh that's good to know. Maybe what we should add then is a case that checks if `val` is a RegExp and extract just the pattern (minus the flags)> We can note in the docstring that cuDF ignores JS RegExp flags.

The goal being that `sr.containsRe(/foo/g)` or `sr.contains(new RegExp('foo'))` work the same as `sr.containsRe('foo')`.

bryevdv: That seems do-able:
```
> r = /foo/ig
/foo/gi
> r.source
'foo'
```
I guess there is some question if there is any other inconsistencies but it looks fairly standard at a glance. I'll proceed with the above.

modules/cudf/test/series/string-tests.ts
```diff
@@ -0,0 +1,54 @@
...
+  'FoO',           // wrong case
+  'bar\n foo',     // multi-line
+  'bar\nfoo'       // multi-line
+];
+
+describe.each([['foo'], [/foo/], [/foo/ig]])('Series regex search (pattern=%p)', (pattern) => {
+  test('containsRe', () => {
+    const expected = [true, true, true, true, true, false, false, true, true];
+    const s        = StringSeries.new(Utf8Vector.from(data));
+    expect([...s.containsRe(pattern).toArrow()]).toEqual(expected);
```
trxcllnt, line 40: nit: Series implements `*[Symbol.iterator]()` method, so we can do this instead:
```suggestion
    expect([...s.containsRe(pattern)]).toEqual(expected);
```

modules/cudf/test/series/string-tests.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  test('countRe', () => {
+    const expected = [1, 1, 1, 2, 1, 0, 0, 1, 1];
+    const s        = StringSeries.new(Utf8Vector.from(data));
+    expect([...s.countRe(pattern).toArrow()]).toEqual(expected);
```
trxcllnt, line 46:
```suggestion
    expect([...s.countRe(pattern)]).toEqual(expected);
```

modules/cudf/test/series/string-tests.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  test('matchesRe', () => {
+    const expected = [true, false, false, true, false, false, false, false, false];
+    const s        = StringSeries.new(Utf8Vector.from(data));
+    expect([...s.matchesRe(pattern).toArrow()]).toEqual(expected);
```
trxcllnt, line 52:
```suggestion
    expect([...s.matchesRe(pattern)]).toEqual(expected);
```

modules/cudf/test/series/string-tests.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2020, NVIDIA CORPORATION.
```
trxcllnt, line 1:
```suggestion
// Copyright (c) 2021, NVIDIA CORPORATION.
```

modules/cudf/src/column/re.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <node_cudf/column.hpp>
+#include <node_cudf/utilities/napi_to_cpp.hpp>
+
+#include <node_rmm/memory_resource.hpp>
+
+#include <cudf/unary.hpp>
+
+#include <rmm/mr/device/per_device_resource.hpp>
+#include "cudf/strings/contains.hpp"
```
trxcllnt, line 23:
```suggestion
#include <cudf/strings/contains.hpp>
```

trxcllnt (approved):
Looks great! Left a few suggestions, but approving anyway.

PR #126, opened 2021-03-18
Add scatter for Series

bryevdv:
Supports scattering single scalars or other series:
```typescript
const col     = Series.new({type: new Int32, data: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]});
const indices = Series.new({type: new Int32, data: [2, 4, 5, 8]});

col.scatter(999, indices);

const values  = Series.new({type: new Int32, data: [200, 400, 500, 800]});
col.scatter(values, indices);
```

Needs a few more tests but ready for a first review. 

Edit: added `setValue`

### DIFF
--- a/modules/cudf/src/node_cudf/table.hpp
+++ b/modules/cudf/src/node_cudf/table.hpp
@@ -172,10 +172,23 @@ class Table : public Napi::ObjectWrap<Table> {
     cudf::size_type threshold,
     rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;
 
+  // table/copying.cpp
   ObjectUnwrap<Table> gather(
     Column const& gather_map,
     cudf::out_of_bounds_policy bounds_policy = cudf::out_of_bounds_policy::DONT_CHECK,
     rmm::mr::device_memory_resource* mr      = rmm::mr::get_current_device_resource()) const;
+
+  ObjectUnwrap<Table> scatter_scalar(
+    std::vector<std::reference_wrapper<const cudf::scalar>> const& source,
+    Column const& indices,
+    bool check_bounds                   = false,
+    rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;
+
+  ObjectUnwrap<Table> scatter_table(
+    Table const& source,
+    Column const& indices,
+    bool check_bounds                   = false,
+    rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;
 
  private:
   static Napi::FunctionReference constructor;
@@ -187,10 +200,13 @@ class Table : public Napi::ObjectWrap<Table> {
   Napi::Value num_columns(Napi::CallbackInfo const& info);
   Napi::Value num_rows(Napi::CallbackInfo const& info);
   Napi::Value select(Napi::CallbackInfo const& info);
-  Napi::Value gather(Napi::CallbackInfo const& info);
   Napi::Value get_column(Napi::CallbackInfo const& info);
   Napi::Value drop_nulls(Napi::CallbackInfo const& info);
   Napi::Value drop_nans(Napi::CallbackInfo const& info);
+  // table/copying.cpp
+  Napi::Value gather(Napi::CallbackInfo const& info);
+  Napi::Value scatter_scalar(Napi::CallbackInfo const& info);
+  Napi::Value scatter_table(Napi::CallbackInfo const& info);
 
   static Napi::Value read_csv(Napi::CallbackInfo const& info);
   Napi::Value write_csv(Napi::CallbackInfo const& info);
--- a/modules/cudf/src/node_cudf/table.hpp
+++ b/modules/cudf/src/node_cudf/table.hpp
@@ -178,13 +178,13 @@ class Table : public Napi::ObjectWrap<Table> {
     cudf::out_of_bounds_policy bounds_policy = cudf::out_of_bounds_policy::DONT_CHECK,
     rmm::mr::device_memory_resource* mr      = rmm::mr::get_current_device_resource()) const;
 
-  ObjectUnwrap<Table> scatter_scalar(
+  ObjectUnwrap<Table> scatter(
     std::vector<std::reference_wrapper<const cudf::scalar>> const& source,
     Column const& indices,
     bool check_bounds                   = false,
     rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;
 
-  ObjectUnwrap<Table> scatter_table(
+  ObjectUnwrap<Table> scatter(
     Table const& source,
     Column const& indices,
     bool check_bounds                   = false,
--- a/modules/cudf/src/series.ts
+++ b/modules/cudf/src/series.ts
@@ -226,7 +226,7 @@ export class AbstractSeries<T extends DataType = any> {
 
   scatter(source: Series<T>|T['scalarType'],
           indices: Series<Int32>|number[],
-          check_bounds?: boolean,
+          check_bounds = false,
           memoryResource?: MemoryResource): void {
     const dst  = new Table({columns: [this._col]});
     const inds = indices instanceof Series ? indices : new Series({type: new Int32, data: indices});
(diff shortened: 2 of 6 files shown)


trxcllnt:
We could also use this to implement `series.setValue(idx, val)` right?

modules/cudf/src/table/copying.cpp
```diff
@@ -28,4 +29,107 @@ ObjectUnwrap<Table> Table::gather(Column const& gather_map,
...
+  CallbackArgs args{info};
+  if (!Column::is_instance(args[0])) {
+    throw Napi::Error::New(info.Env(), "gather selection argument expects a Column");
+  }
+  auto& selection = *Column::Unwrap(args[0]);
+  if (selection.type().id() == cudf::type_id::BOOL8) {
+    return this->apply_boolean_mask(selection)->Value();
+  }
+  return this->gather(selection)->Value();
+}
```
trxcllnt, line 61: Thanks for moving this!

bryevdv:
> We could also use this to implement series.setValue(idx, val) right?

Seems like it would be a one-liner, happy to add to this PR

modules/cudf/src/table/copying.cpp
```diff
@@ -28,4 +29,105 @@ ObjectUnwrap<Table> Table::gather(Column const& gather_map,
...
+
+  if (args.Length() == 2 or (args.Length() >= 2 and args[2].IsUndefined())) {
+    return scatter_scalar(source, indices)->Value();
+  }
+
+  if (args.Length() > 2 and args[2].IsBoolean()) {
+    auto check_bounds = NapiToCPP(args[2]);
+    if (args.Length() == 3) {
+      return scatter_scalar(source, indices, check_bounds)->Value();
+    } else if (args.Length() == 4 and args[2].IsBoolean()) {
```
trxcllnt, line 93: Isn't this additional check redundant?

```suggestion
    } else if (args.Length() == 4) {
```

modules/cudf/src/table/copying.cpp
```diff
@@ -28,4 +29,105 @@ ObjectUnwrap<Table> Table::gather(Column const& gather_map,
...
+      auto* mr = NapiToCPP(info[3]).operator rmm::mr::device_memory_resource*();
+      return scatter_scalar(source, indices, check_bounds, mr)->Value();
+    }
```
trxcllnt, line 96: I think reading an element out of CallbackInfo that doesn't exist returns an empty/undefined `Napi::Value`, in which case our conversion operator will fall back to returning `get_device_memory_resource()`. So I think we can replace this whole block with this:
```suggestion
    auto* mr = NapiToCPP(info[3]).operator rmm::mr::device_memory_resource*();
    return scatter_scalar(source, indices, check_bounds, mr)->Value();
```

modules/cudf/src/table/copying.cpp
```diff
@@ -28,4 +29,105 @@ ObjectUnwrap<Table> Table::gather(Column const& gather_map,
...
+  }
+  NAPI_THROW(Napi::Error::New(Env(),
```
trxcllnt, line 98: Should probably use the `env` of the caller rather than the `env` of the Table instance.
```suggestion
  NAPI_THROW(Napi::Error::New(info.Env(),
```

modules/cudf/src/table/copying.cpp
```diff
@@ -28,4 +29,105 @@ ObjectUnwrap<Table> Table::gather(Column const& gather_map,
...
+    auto check_bounds = NapiToCPP(args[2]);
+    if (args.Length() == 3) {
+      return scatter_table(source, indices, check_bounds)->Value();
+    } else if (args.Length() == 4 and args[2].IsBoolean()) {
+      auto* mr = NapiToCPP(info[3]).operator rmm::mr::device_memory_resource*();
+      return scatter_table(source, indices, check_bounds, mr)->Value();
+    }
+  }
+  NAPI_THROW(Napi::Error::New(
+    Env(), "scatter_table expects a Table, a Column, and optionally a bool and memory resource"));
```
trxcllnt, line 130:
```suggestion
    info.Env(), "scatter_table expects a Table, a Column, and optionally a bool and memory resource"));
```

modules/cudf/src/table/copying.cpp
```diff
@@ -28,4 +29,105 @@ ObjectUnwrap<Table> Table::gather(Column const& gather_map,
...
+  }
+
+  if (args.Length() > 2 and args[2].IsBoolean()) {
+    auto check_bounds = NapiToCPP(args[2]);
+    if (args.Length() == 3) {
+      return scatter_table(source, indices, check_bounds)->Value();
+    } else if (args.Length() == 4 and args[2].IsBoolean()) {
+      auto* mr = NapiToCPP(info[3]).operator rmm::mr::device_memory_resource*();
+      return scatter_table(source, indices, check_bounds, mr)->Value();
+    }
```
trxcllnt, line 127:
```suggestion
    auto* mr = NapiToCPP(info[3]).operator rmm::mr::device_memory_resource*();
    return scatter_table(source, indices, check_bounds, mr)->Value();
```

modules/cudf/src/node_cudf/table.hpp
```diff
@@ -172,11 +172,24 @@ class Table : public Napi::ObjectWrap<Table> {
...
+    Column const& indices,
+    bool check_bounds                   = false,
+    rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;
+
+  ObjectUnwrap<Table> scatter_table(
+    Table const& source,
+    Column const& indices,
+    bool check_bounds                   = false,
+    rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;
+
```
trxcllnt, line 192: Seems like these could be two overloads of `scatter()`, no?

```suggestion
  ObjectUnwrap<Table> scatter(
    std::vector<std::reference_wrapper<const cudf::scalar>> const& source,
    Column const& indices,
    bool check_bounds                   = false,
    rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;

  ObjectUnwrap<Table> scatter(
    Table const& source,
    Column const& indices,
    bool check_bounds                   = false,
    rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;

```

bryevdv: Unless you have strong feelings, I tend to prefer the separate functions to mirror the `libcudf` functions at the low-level (private-ish) `Table` level

modules/cudf/src/series.ts
```diff
@@ -194,6 +195,52 @@ export class AbstractSeries<T extends DataType = any> {
...
+   * @param memoryResource An optional MemoryResource used to allocate the result's device memory.
+   */
+  scatter(values: Series<T>,
+          indices: Series<Int32>|number[],
+          check_bounds?: boolean,
+          memoryResource?: MemoryResource): void;
+
+  scatter(source: Series<T>|T['scalarType'],
+          indices: Series<Int32>|number[],
+          check_bounds?: boolean,
```
trxcllnt, line 229: Should this default to libcudf's default?
```suggestion
          check_bounds = false,
```

trxcllnt (approved):
LGTM, thanks @bryevdv :pray:

AjayThorve (approved):

modules/cudf/src/table/copying.cpp
```diff
@@ -28,4 +29,98 @@ ObjectUnwrap<Table> Table::gather(Column const& gather_map,
...
+  }
+  auto& indices = *Column::Unwrap(args[1]);
+
+  if (args.Length() == 2 or (args.Length() >= 2 and args[2].IsUndefined())) {
+    return scatter(source, indices)->Value();
+  }
+
+  if (args.Length() > 2 and args[2].IsBoolean()) {
+    auto check_bounds = NapiToCPP(args[2]);
+    auto* mr          = NapiToCPP(info[3]).operator rmm::mr::device_memory_resource*();
```
trxcllnt, line 91:
```suggestion
    bool check_bounds = args[2];
    rmm::mr::device_memory_resource* mr = args[3];
```

PR #129, opened 2021-03-23
add static Series.sequence method

bryevdv:
Very basic tests. Happy to make follow-on issue for more tests after GTC


modules/cudf/src/column/filling.cpp
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2020, NVIDIA CORPORATION.
```
bryevdv, line 1:
```suggestion
// Copyright (c) 2021, NVIDIA CORPORATION.
```
must needs fix from wherever I copied this, too

AjayThorve (approved):
awesome! this is very useful

trxcllnt (approved):
LGTM!

PR #137, opened 2021-03-27
Ensure CPM dependencies build (once) in a common location

trxcllnt:
This PR ensures CPM downloads and builds the RAPIDS C++ dependencies into a central location, then found and used for any downstream builds.

An example:
1. `@rapidsai/cudf` calls `CPMFindPackage(cudf)`
   a. `CPMAddPackage` clones `rapidsai/cudf` repo and stores it in `modules/.cache/cpm/cudf`
   b. `libcudf` sources are built as part of the `node_cudf` CMake build
      * `libcudf.so` and related `.o` files are placed in `modules/.cache/build/cudf-build`
4. `@rapidsai/cugraph` calls `CPMFindPackage(cudf)`
   a. `CPMFindPackage` finds `libcudf` in `modules/.cache/build/cudf-build` (built in step 1b.)
      * `@rapidsai/cugraph` does not rebuild `libcudf.so` or any of its objects
5. `@rapidsai/cuspatial` calls `CPMFindPackage(cudf)`
   a. `CPMFindPackage` finds `libcudf` in `modules/.cache/build/cudf-build` (built in step 1b.)
      * `@rapidsai/cuspatial` does not rebuild `libcudf.so` or any of its objects either
6. ???
7. Profit

### DIFF
--- a/modules/core/CMakeLists.txt
+++ b/modules/core/CMakeLists.txt
@@ -18,7 +18,8 @@ cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
 project(node_rapids_core VERSION 0.0.1 LANGUAGES C CXX)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+unset(CMAKE_LIBRARY_OUTPUT_DIRECTORY)
+unset(CMAKE_LIBRARY_OUTPUT_DIRECTORY CACHE)
 
 option(NODE_RAPIDS_USE_CCACHE "Enable caching compilation results with ccache" ON)
 option(DISABLE_DEPRECATION_WARNINGS "Disable warnings generated from deprecated declarations." ON)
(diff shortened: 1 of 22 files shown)


modules/core/CMakeLists.txt
```diff
@@ -18,7 +18,8 @@ cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
 project(node_rapids_core VERSION 0.0.1 LANGUAGES C CXX)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+unset(CMAKE_LIBRARY_OUTPUT_DIRECTORY)
```
robertmaynard, line 21: I am curious on why you need to unset the location of libraries

robertmaynard (approved):

modules/core/CMakeLists.txt
trxcllnt: Oh this is to work around an issue with the command crafted by `cmake-js` -- they explicitly pass `-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=./build/Release`.

Normally this would be fine, but it causes CMake to put `libcudf.so` into the build dir of the first project to build it, instead of the common shared dir (i.e. `modules/.cache/build/cudf-build/libcudf.so`).

This causes subsequent clean-builds to fail in the following way:

1. `cmake-js` deletes `modules/cudf/build/Release` (`node_cudf` is the first project to build `libcudf.so`)
2. `cudf_DIR` is set to the common build dir
   * `CPMFindPackage(cudf)` sees `cudf-config.cmake` etc. and thinks `libcudf` has been "installed"
4. But `libcudf.so` was in `node_cudf`'s build dir, so it fails with a "missing libcudf.so and no rule to make it" error

PR #139, opened 2021-03-30
Add JIT cache path for faster tests in dev, run Jest tests in verbose mode

trxcllnt:
A small quality-of-life improvement for running tests in container dev.

The first cuDF `yarn test` inside a new container instance is very slow because libcudf has to JIT compile all the binaryops kernels, and it's not obvious this is the case because we're running tests with the `jest-silent-reporter` plugin.

* Define `LIBCUDF_KERNEL_CACHE_PATH` env var in containers so JIT compiled kernels are reused across container instances [1]
* Removes the `jest-silent-reporter` + `verbose: false` from Jest configs, run tests in parallel instead of serially [2]

Note: If you want to use the same libcudf kernel cache path for running outside the container, you can set `LIBCUDF_KERNEL_CACHE_PATH=<absolute/path/to/node-rapids>` in your local `.env` file.

1. image
2. 2021-03-30_12-27-10-fast


modules/core/package.json
```diff
@@ -38,7 +38,6 @@
     "@types/jest": "25.2.2",
     "dotenv": "8.2.0",
     "jest": "26.6.3",
-    "jest-silent-reporter": "0.1.2",
```
bryevdv, line 41: Actually very 👍 for this I get anxious when I don't see any test outputs

bryevdv (approved):

PR #140, opened 2021-03-30
Add Bindings Progress Table

ajschmidt8:
This PR creates a table to track the progress of the bindings that we've implemented in `node-rapids`.

In the first column of each table, I've just copied the methods from the corresponding _Methods_ sections on the [cudf doc page](https://docs.rapids.ai/api/cudf/stable/api.html). ~~The :heavy_minus_sign: & :heavy_check_mark: marks in the tables are not yet accurate. I just wanted to get feedback about the format of things before entering the correct values here.~~

**To Do:**

- [x] Solicit feedback on formatting
- [x] Add other library tables (`cugraph`, `cuspatial`, `rmm`)
- [x] Replace placeholder :heavy_minus_sign: & :heavy_check_mark: with correct values
- [x] Link README.md to tables


bryevdv:
A few thoughts:

*  Having  both ➖ & ✔️ is a bit visually oppressive, I think I'd suggest only having a mark like ✅ for the things that are "done" and leaving slots empty/blank otherwise

* I know it will add vertical space, but I just don't think there is enough overlap of numeric series with (string, struct, list) since the latter don't have any of the (many) math operations. Maybe numeric gets its own table, and (string, struct, list) get a combined (three column) table?

ajschmidt8:
> * Having  both heavy_minus_sign & heavy_check_mark is a bit visually oppressive, I think I'd suggest only having a mark like white_check_mark for the things that are "done" and leaving slots empty/blank otherwise

:+1:  agreed.

> * I know it will add vertical space, but I just don't think there is enough overlap of numeric series with (string, struct, list) since the latter don't have any of the (many) math operations. Maybe numeric gets its own table, and (string, struct, list) get a combined (three column) table?

That sounds like a good idea. Initially I had them all separated, but that seemed to take up a lot of vertical space. I'll make this change

bryevdv:
@ajschmidt8 looking good!  What do you think about  removing all the parens for "function calls"? I am not sure what they add in this context

BINDINGS.md
```diff
@@ -0,0 +1,438 @@
...
+
+| cuSpatial                            |         node-rapids         |
+| ------------------------------------ | :-------------------------: |
+| `directed_hausdorff_distance`        |                             |
+| `haversine_distance`                 |                             |
+| `lonlat_to_cartesian`                |                             |
+| `point_in_polygon`                   |    ✅ (`pointInPolygon`)    |
+| `polygon_bounding_boxes`             | ✅ (`polygonBoundingBoxes`) |
+| `polyline_bounding_boxes`            | ✅(`polylineBoundingBoxes`) |
+| `quadtree_on_points`                 |                             |
```
trxcllnt, line 430: This is also done :+1:

BINDINGS.md
```diff
@@ -0,0 +1,438 @@
...
+## cuSpatial
+
+| cuSpatial                            |         node-rapids         |
+| ------------------------------------ | :-------------------------: |
+| `directed_hausdorff_distance`        |                             |
+| `haversine_distance`                 |                             |
+| `lonlat_to_cartesian`                |                             |
+| `point_in_polygon`                   |    ✅ (`pointInPolygon`)    |
+| `polygon_bounding_boxes`             | ✅ (`polygonBoundingBoxes`) |
+| `polyline_bounding_boxes`            | ✅(`polylineBoundingBoxes`) |
```
trxcllnt, line 429: These aren't done, but their `quadtree_` versions below are :pray:

BINDINGS.md
```diff
@@ -0,0 +1,438 @@
...
+| `quadtree_on_points`                 |                             |
+| `points_in_spatial_window`           |                             |
+| `join_quadtree_and_bounding_boxes`   |                             |
```
trxcllnt, line 432: This one is in there too

exactlyallan:
@ajschmidt8 Looking good, thats a lot of features 😓

ajschmidt8:
@trxcllnt, @bryevdv, should be ready for a final review now.

trxcllnt (approved):

AjayThorve (approved):
LGTM!

PR #143, opened 2021-04-05
Add `Series.unique` and `Series.value_counts`

### LINKED ISSUE #132: Add `Series.unique` and `Series.value_counts`
It's currently fairly cumbersome to just get the unique values in a Series, something like:
```
groups = df.groupBy({by: 'foo'}).getGroupd()
[....groups.keys.get('foo').gather(cudf.Series.new({type: new cudf.Int32, data:  groups.offsets})).toArrow()]
```

Python Cudf has [`unique`](https://docs.rapids.ai/api/cudf/stable/api.html?highlight=unique#cudf.core.series.Series.unique) and related [`value_counts`](https://docs.rapids.ai/api/cudf/stable/api.html?highlight=unique#cudf.core.series.Series.value_counts) that we can implement for `node-rapids`

ajschmidt8:
This PR is still a WIP, but it will add `Series.unique` and `Series.value_counts`.

Shout out to @trxcllnt and @AjayThorve for helping me to get started on this :pray:.

Closes #132.


modules/cudf/src/series.ts
```diff
@@ -481,6 +481,15 @@ export class AbstractSeries<T extends DataType = any> {
    * @summary Hook for specialized Series to override when constructing from a C++ Column.
    */
   protected __construct(inp: Column<T>): Series<T> { return Series.new(inp); }
+
+  value_counts(): DataFrame {
```
trxcllnt, line 485:
```suggestion
  value_counts() {
```

modules/cudf/src/column/stream_compaction.cpp
```diff
@@ -59,4 +59,17 @@ Napi::Value Column::drop_nans(Napi::CallbackInfo const& info) {
...
+                                             cudf::null_equality nulls_equal,
+                                             rmm::mr::device_memory_resource* mr) const {
+  std::vector<cudf::size_type> keys{0};
+  auto result = cudf::drop_duplicates(cudf::table_view{{*this}}, keys, keep, nulls_equal, mr);
+  std::vector<std::unique_ptr<cudf::column>> contents = result->release();
+  return Column::New(std::move(contents[0]));
+}
+
+Napi::Value Column::drop_duplicates(Napi::CallbackInfo const& info) {
+  return drop_duplicates();  // unsure how to use NapiToCPP to pass args here
```
AjayThorve, line 72: Something like this might work
```cpp
CallbackArgs args{info};
return drop_duplicates(args[0], args[1], args[2]); //args[0] -> enum cudf::duplicate_keep_option (add inline operator in NapiToCPP), same for args[1] as cudf::null_equality, args[2] is `mr`
```

ajschmidt8: thanks! added in 3c9654d

modules/cudf/src/series.ts
ajschmidt8: from an offline discussion with Paul, removing this return type causes some other type issues that will require some updates to resolve. Paul will open a separate PR for this while I continue work on `Series.unique`

AjayThorve (approved):
LGTM!

modules/cudf/test/cudf-series-test.ts
```diff
@@ -350,3 +350,24 @@ describe.each([new Int32, new Float32, new Float64])('Series.sequence({type=%p,,
...
+
+test('Series.value_counts', () => {
+  const s      = Series.new({type: new Int32, data: [0, 1, 1, 2, 2, 2]});
+  const result = s.value_counts();
+  const count  = [...result.count.toArrow()];
+  const value  = [...result.value.toArrow()];
+
+  expect(count[value.findIndex((el) => el === 0)]).toBe(1);
+  expect(count[value.findIndex((el) => el === 1)]).toBe(2);
+  expect(count[value.findIndex((el) => el === 2)]).toBe(3);
```
bryevdv, line 362: I guess the order is not guaranteed? would it be possible to sort `result` first so that `count` and `value` can just be compared to concrete expected values?

modules/cudf/src/types/mappings.ts
```diff
@@ -44,6 +44,12 @@ export enum Interpolation
   nearest    ///< i or j, whichever is nearest
 }
 
+export enum NullEquality
+{
+  equal,    // nulls compare equal
+  unequal,  // nulls compare unequal
+}
```
bryevdv, line 51: FWIW in the joins PR I was just going to use a `nullEquality` boolean in places, since there are only two values and that seems less of a burden on users. But mainly we should just be consistent either way. @trxcllnt do you have a preference?

modules/cudf/src/column/stream_compaction.cpp
```diff
@@ -59,4 +64,17 @@ Napi::Value Column::drop_nans(Napi::CallbackInfo const& info) {
   return drop_nans(NapiToCPP(info[0]).operator rmm::mr::device_memory_resource*());
 }
 
+ObjectUnwrap<Column> Column::drop_duplicates(cudf::null_equality nulls_equal,
+                                             rmm::mr::device_memory_resource* mr) const {
+  return Column::New(std::move(
+    cudf::drop_duplicates(
+      cudf::table_view{{*this}}, {0}, cudf::duplicate_keep_option::KEEP_FIRST, nulls_equal, mr)
+      ->release()[0]));
+}
```
bryevdv, line 73: I think lately the practice has been to wrap libcudf calls in `try...catch` to convert any c++ exceptions into  `NAPI_ERROR` e.g. 

https://github.com/rapidsai/node-rapids/blob/95c80d6cf655b8c6e7b33d662c2f3356850e4edd/modules/cudf/src/column/filling.cpp#L47-L56

but I also notice the existing functions don't seem to. @trxcllnt should this error-handling pattern be extended here?

trxcllnt: Yeah we should wrap the calls. If libcudf throws an error here, it'll lead to process abort unless we catch and rethrow as a JS exception.

modules/cudf/test/cudf-series-test.ts
ajschmidt8: Yeah, that seems like a better approach. Do you have any thoughts on how this might be implemented? If `value_counts` returned a DataFrame, it seems like `sortValues` may work. But since `value_counts` returns an object ([src](https://github.com/rapidsai/node-rapids/blob/b2a14bb3bef071cb3a4431c90c45d0317fb85499/modules/cudf/src/series.ts#L503-L506)) with 2 separate series, I'm not sure the best way to go about this. Paul and I tried to make `value_counts` return a DataFrame, but we ran into issues with circular references in TypeScript since a DataFrame consists of Series.

modules/cudf/src/types/mappings.ts
ajschmidt8: :+1:. Unless Paul disagrees, I'm fine with switching to a boolean here.

modules/cudf/test/cudf-series-test.ts
bryevdv: TBH I don't even think the sort is required? just iterating over the indices [0, 1, 2] would do. As another suggestion: maybe change the values in the original data to be things that are clearly neither indices, nor counts? e.g. 
```
Series.new({type: new Int32, data: [1000, 1100, 1100, 1200, 1200, 1200]});
```
(Probably would not hurt to "un-sort" those values too, unless that is an assumption of the method)

modules/cudf/src/column/stream_compaction.cpp
ajschmidt8: updated in c6feecc

modules/cudf/test/cudf-series-test.ts
ajschmidt8: updated in 239f39a

modules/cudf/src/column/stream_compaction.cpp
bryevdv: As long as this file is being modified I think it would be good to update the other functions above that are missing `try...catch` at the same time. But if not let's make an issue about it so it is not forgotten.

modules/cudf/src/types/mappings.ts
trxcllnt: I think boolean here is fine! We use enums in C++ because calling a method with `cudf::blah(true, false, true)` is super weird, but we're using bag-o-options in JS, so it's obvious from the key name what the boolean means.

modules/cudf/src/node_cudf/column.hpp
```diff
@@ -16,6 +16,7 @@
 
 #include <node_cudf/scalar.hpp>
 #include <node_cudf/utilities/dtypes.hpp>
+#include "cudf/stream_compaction.hpp"
```
trxcllnt, line 19:
```suggestion

#include <cudf/stream_compaction.hpp>
```

modules/cudf/src/node_cudf/utilities/napi_to_cpp.hpp
```diff
@@ -16,6 +16,7 @@
 
 #include <node_cudf/scalar.hpp>
 #include <node_cudf/utilities/dtypes.hpp>
+#include "cudf/stream_compaction.hpp"
```
trxcllnt, line 19:
```suggestion

#include <cudf/stream_compaction.hpp>
```

modules/cudf/src/types/mappings.ts
trxcllnt: Though if we're not using option-bag arguments, then the enum is probably the way to go (or `keyof` an enum like we do in [other places](https://github.com/rapidsai/node-rapids/blob/017cfb718298aead959f8e6cfbeb02c786af051b/modules/cudf/src/series/numeric.ts#L1401-L1406)).

trxcllnt (changes requested):
Made a few small suggestions. Other than that, looks good to me!

modules/cudf/test/cudf-series-test.ts
```diff
@@ -350,3 +350,28 @@ describe.each([new Int32, new Float32, new Float64])('Series.sequence({type=%p,,
...
+    const currentCount = count[i];
+    expect(currentCount).toBe(countMap[currentVal]);
+  }
+});
+
+test.each`
+nulls_equal        | data                           | expected
+${'equal'}         | ${[null, null, 1, 2, 3, 4, 4]} | ${[null, 1, 2, 3, 4]}
+${'unequal'}       | ${[null, null, 1, 2, 3, 4, 4]} | ${[null, null, 1, 2, 3, 4]}
+`('Series.unique($nulls_equal)', ({nulls_equal, data, expected}) => {
```
trxcllnt, line 373: I like this very much!

modules/cudf/src/types/mappings.ts
ajschmidt8: updated in 8d19015

trxcllnt (approved):

PR #144, opened 2021-04-07
Update default TypeScript Installation

ajschmidt8:
This PR adds a `typescript.tsdk` entry to the workspace file, which sets the default TypeScript installation that should be used by VS Code in the workspace.


trxcllnt (approved):
:pray:

PR #163, opened 2021-05-03
[2/9] Add new EnvLocalAddon and EnvLocalObjectWrap helper classes

trxcllnt:
See https://github.com/rapidsai/node-rapids/pull/171 for more info.

tl;dr:

* Define helpers that store and load class constructors in the local `Napi::Env`'s instance data.
  * This allows us to avoid using static FunctionReferences, which runs the risk of creating JS instances using an Napi::Env other than the currently active env.
* Remove code made obsolete by these two new helper classes
* Light style cleanup

#### The problem

The new blessed N-API way to build native addons is by extending the [`Napi::Addon`](https://github.com/nodejs/node-addon-api/blob/main/doc/addon.md) class.

Instead of defining a top-level `initModule(Napi::Env env, Napi::Object exports)` function and attaching things to `exports`... we extend a class with a constructor that takes `(Napi::Env env, Napi::Object exports)` and attaches things to `exports` :joy:.

If that was the only difference, this change wouldn't have been worth it. The `Napi::Addon` class also [sets itself as the global instance data](https://github.com/nodejs/node-addon-api/blob/93f189831253e392bddc71c409ad61ea1024626e/napi-inl.h#L5631-L5632) of the addon's [`napi_env`](https://nodejs.org/api/n-api.html#n_api_napi_env). Each native addon is run in its own `napi_env`. The `napi_env` is used by the runtime to track VM-specific state.

Most N-API examples define the `Napi::FunctionReference` for C++-defined constructors as global static variables. Both the `Function` and `FunctionReference` were created in the context of a particular `napi_env`, so invoking these constructors in another `napi_env` is problematic.

Furthermore, most N-API examples out there _don't_ dynamically link to other native addons, so there aren't robust examples we can mimic. We've been able to get away with global statics so far because libs loaded in the same Linux process share the global symbol table (this is not the case in Windows).

#### A solution

There may be more than one way to solve this, but the one that seemed easiest was to manage our own "symbol table" (of sorts) as part of the `napi_env` global instance data.

The [`std::type_info` provides a unique hashcode](https://en.cppreference.com/w/cpp/types/type_info/hash_code) for each C++ type that we can use as the key to store each C++ class's `Napi::Function` constructor in a JS object:

```C++
#include <typeinfo>

template <typename Class>
inline Napi::Function InitClass(Napi::Env const& env, Napi::Object exports) {
  auto const constructor = Class::Init(env, exports);
  env.GetInstanceData<EnvLocalAddon>()->_cpp_exports.Set(
    typeid(Class).hash_code(), constructor
  );
  return constructor;
}
```

When we want to instantiate an instance of a class, we can look up its constructor function by name in the `napi_env`'s global instance data (i.e. the top-level `Napi::Addon`):

```C++
template <typename Class>
inline Napi::Function GetConstructor(Napi::Env const& env) {
  return env.GetInstanceData<EnvLocalAddon>()->_cpp_exports.Get(
    typeid(Class).hash_code()
  ).As<Napi::Function>();
}

// Look ma, no statics!
auto inst = GetConstructor<MyClass>(env).New({ /*...args*/ });
```

To automate/facilitate this pattern, this PR adds two new classes that handle all this for us automatically: `EnvLocalAddon`, and `EnvLocalObjectWrap`.

#### `EnvLocalObjectWrap`: a friendlier `Napi::ObjectWrap`

In addition to automatically looking up function constructors in the global instance data map, `EnvLocalObjectWrap` makes it easier to bridge the gap between the instance of our C++ class and the `Napi::Object` wrapper whose lifetime determines the lifetime of our C++ instance.

The constructor function for a regular `ObjectWrap` returns a type-erased `Napi::Object`:

```C++
struct MyClass : public Napi::ObjectWrap<MyClass> { get_5() { return 5; }};
Napi::Function MyClassCtor = MyClass::DefineClass("MyClass", {});

 // ew!
Napi::Object wrapper = MyClassCtor.New(env, {});
if (wrapper.InstanceOf(MyClass)) { // gross!
  MyClass* instance = MyClass::Unwrap(wrapper); // yuck!
  std::cout << instance->get_5() << std::endl;
}
```

So instead of the above, `EnvLocalObjectWrap` has a special type that extends `Napi::Object` and provides access to the underlying C++ type more easily:

```C++
struct MyClass : public EnvLocalObjectWrap<MyClass> {
  inline static Init(Napi::Env env, Napi::Object exports) { return DefineClass("MyClass", {}); }
  get_5() { return 5; }
};
// Create and register MyClass's constructor function
env.GetInstanceData<EnvLocalAddon>()->InitClass<MyClass>(env, exports);

// strongly typed!
MyClass::wrapper_t wrapper = MyClassCtor.New(env, {});
// Access members of underlying C++ instance
std::cout << wrapper->get_5() << std::endl;

// Get reference to underlying C++ instance
MyClass const& inst = *wrapper;
std::cout << inst.get_5() << std::endl;

// Or implicitly convert MyClass::wrapper_t to MyClass&
[](MyClass const& inst) {
  std::cout << inst.get_5() << std::endl;
}(wrapper);
```

This utility means we can use the JS `MyClass::wrapper_t` instance (whose lifetime we'd have to manage anyway), and eliminate all the ceremony around unwrapping, and avoid the bugs from allowing the JS object to be deleted and taking down our C++ instance with it.


AjayThorve (approved):
thanks @trxcllnt for the explanation behind the changes, makes sense!

PR #167, opened 2021-05-03
[6/9] Update cudf to use new EnvLocalAddon and EnvLocalObjectWrap

trxcllnt:
See https://github.com/rapidsai/node-rapids/pull/171 for more info.

* Update `@rapidsai/cudf` module to use new `EnvLocalAddon` and `EnvLocalObjectWrap` helper classes
* Make cuDF use a larger file cache for compiled Jitify kernels
* Light cleanup


modules/cudf/src/column/copying.cpp
```diff
@@ -21,12 +21,13 @@
...
-  auto result = cudf::gather(cudf::table_view{{*this}}, gather_map, bounds_policy, mr);
-  std::vector<std::unique_ptr<cudf::column>> contents = result->release();
-  return Column::New(std::move(contents[0]));
+Column::wrapper_t Column::gather(Column const& gather_map,
+                                 cudf::out_of_bounds_policy bounds_policy,
+                                 rmm::mr::device_memory_resource* mr) const {
+  return Column::New(
+    Env(),
+    std::move(
+      cudf::gather(cudf::table_view{{*this}}, gather_map, bounds_policy, mr)->release()[0]));
```
bryevdv, line 30: What's the rationale for combining things into one giant expression instead of smaller individual steps?

modules/cudf/src/column/reduction.cpp
```diff
@@ -46,86 +47,98 @@ cudf::data_type _compute_dtype(cudf::type_id id) {
...
 
-ObjectUnwrap<Scalar> Column::mean(rmm::mr::device_memory_resource* mr) const {
+Scalar::wrapper_t Column::mean(rmm::mr::device_memory_resource* mr) const {
   return reduce(cudf::make_mean_aggregation(), cudf::data_type(cudf::type_id::FLOAT64), mr);
 }
 
 Napi::Value Column::mean(Napi::CallbackInfo const& info) {
-  return mean(NapiToCPP(info[0]).operator rmm::mr::device_memory_resource*());
+  return Napi::Value::From(info.Env(),
+                           mean(NapiToCPP(info[0]).operator rmm::mr::device_memory_resource*()));
```
bryevdv, line 129: I noticed some of the `_re` functions above switched to using `CallbackArgs args{info}` and avoided the explicit 
```
NapiToCPP(info[0]).operator rmm::mr::device_memory_resource
```
As magic goes, that's one bit I could live with, is there a reason not to use `CallbackArgs args{info}` more liberally (or even conventionally)?

modules/cudf/src/column/copying.cpp
trxcllnt: Just trying to make things more consistent while I was find-replacing all the `ObjectUnwrap` return signatures. We can make things more consistent the other way, but I like using temporaries instead of lvalues.

modules/cudf/src/column/reduction.cpp
trxcllnt: Yeah, I can see an argument to be made for consistency here. I usually only construct a `CallbackArgs` if I need to convert multiple arguments, i.e.:

```c++
CallbackArgs const args{info}
foo(args[0], args[1], args[2]);
```

But I've also used it as a shortcut for a single argument before:

```c++
bar(CallbackArgs{info}[0].operator rmm::mr::device_memory_resource*());
// vs.
bar(NapiToCPP{info[0]}.operator rmm::mr::device_memory_resource*());
```

I like the idea of only using `CallbackArgs`, and making `NapiToCPP` an internal implementation detail of `CallbackArgs`. I'll also do some tests to see if I can make `CallbackArgs` a drop-in replacement for `Napi::CallbackInfo` so we can use it in methods defined via the ObjectWrap APIs.

modules/cudf/src/column/copying.cpp
bryevdv: It just gets difficult to visually parse what is what when there are nested function calls, derefs and array indexes in a single expression

PR #175, opened 2021-05-06
Added series nonnull binding

mzegar:
**Changes**
- Added `isNotNull` test to `cudf-series-test.ts`
- Renamed `isValid` to `isNotNull`
   - In `float.ts` we have `isNotNull` and `isNotNaN` already implemented. It makes a bit more sense to use `isNotNull` instead of `isValid` to differentiate between `isNotNull` and `isNotNaN`. [The python bindings do the same thing.](https://github.com/rapidsai/cudf/blob/8ae73d50cee62a282cba152cbce7d23449b5a179/python/cudf/cudf/core/column/column.py#L757) 
- Similar to this [PR](https://github.com/rapidsai/node-rapids/pull/174), marked off any instances of `notna` with `isNotNull` since `notna` only exists on floating point types.

### DIFF
--- a/BINDINGS.md
+++ b/BINDINGS.md
@@ -58,8 +58,8 @@ The tables below show the bindings that have been implemented in `node-rapids`.
 | `head`               |                     |
 | `interleave_columns` |                     |
 | `isin`               |                     |
-| `isna`               |                     |
-| `isnull`             |                     |
+| `isna`               |    ✅ (`isNull`)    |
+| `isnull`             |         ✅          |
 | `keys`               |                     |
 | `kurt`               |                     |
 | `kurtosis`           |                     |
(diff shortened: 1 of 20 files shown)


trxcllnt (changes requested):
Renaming `isValid` to `isNotNull` is fine with me, but looks like the [unaryops tests](https://github.com/rapidsai/node-rapids/blob/main/modules/cudf/test/series/unaryop/float32-tests.ts#L54-L57) need to be updated.

mzegar:
> Renaming `isValid` to `isNotNull` is fine with me, but looks like the [unaryops tests](https://github.com/rapidsai/node-rapids/blob/main/modules/cudf/test/series/unaryop/float32-tests.ts#L54-L57) need to be updated.

Good catch, will change.

AjayThorve:
> Any reason why github is showing the changes made by @trxcllnt #176 in my change list even though they have been merged in? Is it because I'm working off a fork?

This could be because you created a branch off `main` before #176 was merged and later merged that (same point that it shows up here in logs). If you don't want it to show up, you can always try `rebase`, but honestly, this should be fine as it is.

mzegar:
> > Any reason why github is showing the changes made by @trxcllnt #176 in my change list even though they have been merged in? Is it because I'm working off a fork?
> 
> This could be because you created a branch off `main` before #176 was merged and later merged that (same point that it shows up here in logs). If you don't want it to show up, you can always try `rebase`

Found this... 

https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

Simple enough I guess.

trxcllnt (approved):
LGTM :+1:

PR #176, opened 2021-05-06
Clean up tests to use spread enumeration

trxcllnt:
PR with a bit of cleanup in the tests. `[...ser.toArrow()]` is redundant now that we have the `Series[Symbol.iterator]()` method, because it internally calls `this.toArrow()`.


trxcllnt:
cc @mzegar

AjayThorve (approved):
LGTM! internally calling toArrow() makes it super convenient!

PR #182, opened 2021-05-10
[FEA] Add `Series` concat/append

mzegar:
Adds in the `Series` `concat` mentioned in https://github.com/rapidsai/node-rapids/issues/97

**Changes**
- Added `concat` for the `Series` type
- Added `concat` tests for `Numeric`, `List`, `String`, and `Struct`

**Unknowns**
- Not sure if additional handling / error checking needs to occur in `concat` @ `series.ts`
- If the param should be called `other` instead of `rhs` (or maybe other ideas?) `concat` @ `series.ts`


modules/cudf/src/column/concatenate.cpp
```diff
@@ -0,0 +1,35 @@
...
+    return Column::New(Env(), cudf::concatenate(std::vector{this->view(), other}, mr));
+  } catch (std::exception const& e) { NAPI_THROW(Napi::Error::New(Env(), e.what())); }
+}
+
+Napi::Value Column::concat(Napi::CallbackInfo const& info) {
+  CallbackArgs args{info};
+  try {
+    return concat(*Column::Unwrap(args[0]), args[1]);
+  } catch (cudf::logic_error const& e) { NAPI_THROW(Napi::Error::New(info.Env(), e.what())); }
+  throw Napi::Error::New(info.Env(), "Concat requires a Column");
```
bryevdv, line 32: This logic seems a little off, since public `concat` already catches all exceptions. Even if it didn't, how would the last line be reach (if there were still some other uncaught exception after the `catch` it would just keep propagating up the stack, no?)

Recent pattern seems more along these lines:

* catch `catch (cudf::logic_error const& e)` in the public `concat` that actually calls `cudf` functions
* don't wrap the call to public `concat` in any `try...except`
* instead, do explicit argument number/type checking to raise "Concat requires a Column" before calling public `concat` if necessary

cc @trxcllnt

mzegar: Yup, the throw statement is completely useless in this case (for whatever reason assumed that the throw statement was apart of the `try...catch` block above).

and I'm pretty sure based off of this method definition
https://github.com/rapidsai/node-rapids/blob/ebc27db6c44cfa16d07e810c3712d7a679bf6ab2/modules/cudf/src/series.ts#L438

It shouldn't be possible to call it without a `Column`, so I don't think I even need any argument checking. (let me know if this isn't the case).

Thanks for the review.

modules/cudf/src/series.ts
```diff
@@ -419,6 +419,28 @@ export class AbstractSeries<T extends DataType = any> {
...
+   * // Float64Series
+   * Series.new([1, 2, 3]).concat(Series.new([4, 5, 6])) // [1, 2, 3, 4, 5, 6]
+   * // StringSeries
+   * Series.new(["foo"]).concat(Series.new(["bar"])) // ["foo", "bar"]
+   *
+   * ```
+   */
+  concat(other: Series<T>, memoryResource?: MemoryResource): Series<T>;
+
+  concat(other: any, memoryResource?: MemoryResource): Series<T> {
```
bryevdv, line 440: Why is the second `other: any` signature needed?

mzegar: IDE was complaining about passing in `other._col` when the method definition looks like this...
```
  concat(other: Series<T>, memoryResource?: MemoryResource): Series<T> {
    return Series.new(this._col.concat(other._col, memoryResource));
  }
 ```
 
 ```
 Argument of type 'Column<Int64> | Column<Float64> | Column<Bool8> | Column<Int32> | Column<Int8> | Column<Int16> | ... 7 more ... | Column<...>' is not assignable to parameter of type 'Column<T>'.
  Type 'Column<Int64>' is not assignable to type 'Column<T>'.
    Type 'Int64' is not assignable to type 'T'.
      'Int64' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'DataType'.ts(2345)
 ```
 
 It seems like `Int64` is apart of `DataType`, so this is a bit odd...

Comment (mzegar) [reply modules/cudf/src/series.ts:440]: Spoke with @bryevdv a bit about this, ended up with just casting `other._col`
```
return Series.new(this._col.concat(other._col as Column<T>, memoryResource));
```
and removing the `other: any` signature. Let me know if you have any other ideas @trxcllnt , otherwise I'll update to use this.

Comment (trxcllnt) [reply modules/cudf/src/series.ts:440]: That fix looks correct, unfortunately... :confused:

**Review by trxcllnt — COMMENTED** [review]
Looks good to me so far. Should we add the [dtype upcasting](https://github.com/rapidsai/cudf/blob/c2c67de0a487ef767f2b16e6b95132121e2eec04/python/cudf/cudf/core/series.py#L2404-L2408) to this PR as well?

The easiest way I can think to do it is to implement `concat()` on the individual subclasses, that way the child classes' `concat` signatures aren't restricted in the types they can return:

```typescript
// series/numeric.ts
import { CUDF } from '../addon';
import { CommonType } from '../types/mappings.ts';

class NumericSeries<T extends Numeric> extends Series {
  concat<R extends Numeric>(other: Series<R>, memoryResource?: MemoryResource): Series<CommonType<T, R>> {
    const type = CUDF.findCommonType(this.type, other.type);
    return Series.new(
      this._col.cast(type, memoryResource).concat(
        other._col.cast(type, memoryResource),
        memoryResource
      ));
  }
}

// series/string.ts
class StringSeries extends Series {
  concat(other: Series<Utf8String>, memoryResource?: MemoryResource): Series<Utf8String> {
    return this.__construct(this._col.concat(other._col, memoryResource));
  }
}

// ...etc.
```

modules/cudf/src/column/concatenate.cpp
mzegar: https://github.com/rapidsai/node-rapids/pull/183#discussion_r629810222

Only using `try...catch` inside of the `libcudf` call, removing other instance.

bryevdv (approved):

mzegar:
> Looks good to me so far. Should we add the [dtype upcasting](https://github.com/rapidsai/cudf/blob/c2c67de0a487ef767f2b16e6b95132121e2eec04/python/cudf/cudf/core/series.py#L2404-L2408) to this PR as well?
> 
> The easiest way I can think to do it is to implement `concat()` on the individual subclasses, that way the child classes' `concat` signatures aren't restricted in the types they can return:
> 
> ```ts
> // series/numeric.ts
> import { CUDF } from '../addon';
> import { CommonType } from '../types/mappings.ts';
> 
> class NumericSeries<T extends Numeric> extends Series {
>   concat<R extends Numeric>(other: Series<R>, memoryResource?: MemoryResource): Series<CommonType<T, R>> {
>     const type = CUDF.findCommonType(this.type, other.type);
>     return Series.new(
>       this._col.cast(type, memoryResource).concat(
>         other._col.cast(type, memoryResource),
>         memoryResource
>       ));
>   }
> }
> 
> // series/string.ts
> class StringSeries extends Series {
>   concat(other: Series<Utf8String>, memoryResource?: MemoryResource): Series<Utf8String> {
>     return this.__construct(this._col.concat(other._col, memoryResource));
>   }
> }
> 
> // ...etc.
> ```

Thanks for these code snippets, pushed my changes.

AjayThorve:
> ```ts
> // series/numeric.ts
> import { CUDF } from '../addon';
> import { CommonType } from '../types/mappings.ts';
> 
> class NumericSeries<T extends Numeric> extends Series {
>   concat<R extends Numeric>(other: Series<R>, memoryResource?: MemoryResource): Series<CommonType<T, R>> {
>     const type = CUDF.findCommonType(this.type, other.type);
>     return Series.new(
>       this._col.cast(type, memoryResource).concat(
>         other._col.cast(type, memoryResource),
>         memoryResource
>       ));
>   }
> } 
> ```

just a thought, @trxcllnt , shouldn't the type of current series take precedence instead of finding a common type for returned series?

bryevdv:
@AjayThorve is this the analogue of `append` in Python? If so it takes the common type either way:
```
In [2]: df = cudf.DataFrame([[1, 2], [3, 4]], columns=list('AB'))

In [3]: df2 = cudf.DataFrame([[5.0, 6.0], [7, 8]], columns=list('AB'))

In [4]: df.append(df2)
Out[4]: 
     A    B
0  1.0  2.0
1  3.0  4.0
0  5.0  6.0
1  7.0  8.0

In [5]: df2.append(df)
Out[5]: 
     A    B
0  5.0  6.0
1  7.0  8.0
0  1.0  2.0
1  3.0  4.0
```

AjayThorve (approved):

modules/cudf/src/series/numeric.ts
```diff
@@ -78,6 +79,25 @@ export abstract class NumericSeries<T extends Numeric> extends Series<T> {
...
+   * import {Series} from '@rapidsai/cudf';
+   *
+   * Series.new([1, 2, 3]).concat(Series.new([4, 5, 6])) // [1, 2, 3, 4, 5, 6]
+   * ```
+   */
+  concat<R extends Numeric>(other: Series<R>,
+                            memoryResource?: MemoryResource): Series<CommonType<T, R>> {
+    const type = CUDF.findCommonType(this.type, other.type);
+    return Series.new(this._col.cast(type, memoryResource)
+                        .concat(other._col.cast(type, memoryResource), memoryResource));
```
trxcllnt, line 98: Small performance optimization suggestion -- we can avoid one or both of the cast ops if the dtypes are already the same as what we're casting to. This leverages the fact that Arrow's `DataType` classes have a `compareTo` method for checking dtype equivalence:

```suggestion
    const type = CUDF.findCommonType(this.type, other.type);
    const lhs = type.compareTo(this.type) ? this._col : this._col.cast(type, memoryResource);
    const rhs = type.compareTo(other.type) ? other._col : other._col.cast(type, memoryResource);
    return Series.new(lhs.concat(rhs, memoryResource));
```

trxcllnt (changes requested):
Looks good! Just left a few small suggestions.

trxcllnt (approved):
Resolved the conflicts, committed suggestions, and tested locally. Works on my machine :wink:!

PR #187, opened 2021-05-11
[FEA] Create `ListSeries` from JS Arrays, add `ListSeries.getValue()`

trxcllnt:
Now that [`cudf::get_element` supports `list_scalar`](https://github.com/rapidsai/cudf/pull/8071), we can add `ListSeries.getValue()`.

* Support creating `ListSeries` from Arrays of JS Arrays
* Move `getValue` and `setValue` to the Series subclasses to make the types happy
* Implement `ListSeries.getValue()`

### DIFF
--- a/modules/cudf/src/series.ts
+++ b/modules/cudf/src/series.ts
@@ -534,51 +534,6 @@ export class AbstractSeries<T extends DataType = any> {
   filter(mask: Series<Bool8>): Series<T> { return this.__construct(this._col.gather(mask._col)); }
 
   /**
-   * Return a value at the specified index to host memory
-   *
-   * @param index the index in this Series to return a value for
-   *
-   * @example
-   * ```typescript
-   * import {Series} from "@rapidsai/cudf";
-   *
-   * // Float64Series
-   * Series.new([1, 2, 3]).getValue(0) // 1
-   * // StringSeries
-   * Series.new(["foo", "bar", "test"]).getValue(2) // "test"
-   * // Bool8Series
-   * Series.new([false, true, true]).getValue(3) // throws index out of bounds error
-   * ```
-   */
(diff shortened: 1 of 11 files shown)


modules/cudf/src/series.ts
```diff
@@ -875,21 +893,72 @@ export {
...
+  if (value.length === 0) { return new Float64; }
+  let nullsCount    = 0;
+  let arraysCount   = 0;
+  let objectsCount  = 0;
+  let numbersCount  = 0;
+  let stringsCount  = 0;
+  let bigintsCount  = 0;
+  let booleansCount = 0;
+  let unknownCount  = 0;
+  value.forEach((val) => {
```
AjayThorve, line 905: Much more efficient, since it only loops over the values once!

AjayThorve (approved):

modules/cudf/src/series/list.ts
```diff
@@ -74,6 +75,30 @@ export class ListSeries<T extends DataType> extends Series<List<T>> {
...
+   * // Series<List<Utf8String>>
+   * Series.new([["foo", "bar"], ["test"]]).getValue(1) // Series(["test"])
+   *
+   * // Series<List<Bool8>>
+   * Series.new([[false, true], [true]]).getValue(2) // throws index out of bounds error
+   * ```
+   */
+  getValue(index: number) {
+    const value = this._col.getValue(index);
+    return value === null ? null : Series.new(value);
```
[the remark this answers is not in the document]
trxcllnt: All columns can return null for an element, i.e.
```js
Series.new([null, [1, 2], null]).getValue(0) // null
```

modules/cudf/src/series.ts
trxcllnt: @bryevdv yes I believe it could!

PR #188, opened 2021-05-11
[FEA] Added `isNull` to `dataframe`

matekdev:
~~Wanted to take on an easier dataframe operation first before jumping into `concat`. Marking as draft as I have a few dataframe specific questions I'd appreciate answered.~~

**Changes**
- Added `isNull` for `DataFrame`
- Added `isNull` test in `cudf-data-frame-test`
- Changed float typechecking in `DataFrame`, no longer uses `instanceof`

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -323,11 +323,7 @@ export class DataFrame<T extends TypeMap = any> {
    * @param memoryResource The optional MemoryResource used to allocate the result Series's device
    *   memory.
    * @returns DataFrame of Series cast to the new dtype
-   *
-   * @example
-   * ```typescript
-   * import {DataFrame, Series, Int32, Float32}  from '@rapidsai/cudf';
-   * const df = new DataFrame({
+   *make notebooks.run
    *  a: Series.new({type: new Int32, data: [0, 1, 1, 2, 2, 2]}),
    *  b: Series.new({type: new Int32, data: [0, 1, 2, 3, 4, 4]})
    * })
@@ -907,9 +903,8 @@ export class DataFrame<T extends TypeMap = any> {
    * ```
    */
   isNull(): DataFrame<{[P in keyof T]: Bool8}> {
-    return new DataFrame(new ColumnAccessor(this.names.reduce((cols, name) => {
-      cols[name] = this.get(name).isNull()._col;
-      return cols;
-    }, {} as ColumnsMap<{[P in keyof T]: Bool8}>)));
+    return new DataFrame(
+      this.names.reduce((cols, name) => ({...cols, [name]: this.get(name).isNull()}),
+                        {} as SeriesMap<{[P in keyof T]: Bool8}>));
   }
 }


modules/cudf/src/data_frame.ts
```diff
@@ -884,4 +884,35 @@ export class DataFrame<T extends TypeMap = any> {
...
+   * import {DataFrame, Series, Int32, Float32}  from '@rapidsai/cudf';
+   * const df = new DataFrame({
+   *  a: Series.new({type: new Int32, data: [0, 1, 2, 3, 4, 4]}),
+   *  b: Series.new({type: new Float32, data: [0, NaN, 2, 3, 4, 4]})
+   * });
+   *
+   * df.isNull()
+   * ```
+   */
+  isNull(): DataFrame<T> {
```
mzegar, line 906: I saw that `subset` was an optional param for a lot of the other dataframe operations, do we want to include this for basically all dataframe methods (where it makes sense?)

modules/cudf/src/data_frame.ts
```diff
@@ -884,4 +884,35 @@ export class DataFrame<T extends TypeMap = any> {
...
+    const temp       = new Table({columns: this.select(this.names)._accessor.columns});
+    const series_map = {} as SeriesMap<T>;
+    this._accessor.names.forEach((name, index) => {
+      if (this.get(name) instanceof Float32Series || this.get(name) instanceof Float64Series) {
+        series_map[name] = Series.new(temp.getColumnByIndex(index).isNaN()) as any;
```
mzegar, line 911: I am forced to cast this as `any` as I get this IDE error otherwise

```
Type 'Bool8Series' is not assignable to type 'AbstractSeries<T[string & keyof T]>'.
  Types of property '_col' are incompatible.
    Type 'Column<Bool8>' is not assignable to type 'Column<T[string & keyof T]>'.
      Type 'Bool8' is not assignable to type 'T[string & keyof T]'.ts(2322)
```

I feel like this could be fixed by extending the `Bool8Series` type into another one of the parameters, but, my knowledge of how types inherit in this app is a bit lacking, any suggestions?

modules/cudf/src/data_frame.ts
trxcllnt: it doesn't look like Python's [`DataFrame.isnull()`](https://github.com/rapidsai/cudf/blob/2c70f1d5084c12bd04f676b95d5f20cb53f67250/python/cudf/cudf/core/frame.py#L2264-L2339) accepts a subset argument, so we shouldn't need one here. 

The `dropNulls` and `dropNaN` methods accept a subset, i.e. a list of column names to consider when dropping rows. This allows the user to exclude certain columns from consideration when dropping rows that contain null/NaN values:

```js
console.log(...new DataFrame({
  a: Series.new([1, null, 2]),
  b: Series.new([null, 1, 2]),
}).dropNulls(1, 1, ['a']).toArrow()]);
// [
//   { a: 1, b: null },
//   { a: 2, b: 2 }
// ]

```

mzegar: Ah, the one example I looked at was `nansToNulls` which is defined above, it takes in a `subset` even though the docs mention it shouldn't. Shall I remove the arg?
https://docs.rapids.ai/api/cudf/nightly/api.html#cudf.core.dataframe.DataFrame.nans_to_nulls

modules/cudf/src/data_frame.ts
```diff
@@ -884,4 +884,39 @@ export class DataFrame<T extends TypeMap = any> {
...
+      } else {
+        series_map[name] = Series.new(temp.getColumnByIndex(index).isNull()) as any;
+      }
+    });
+    return new DataFrame(series_map);
+  }
```
trxcllnt, line 921: Here's how I'd implement this:

```suggestion
  public isNull() {
    return new DataFrame(new ColumnAccessor(this.names.reduce((cols, name) => {
      cols[name] = this.get(name).isNull()._col;
      return cols;
    }, {} as ColumnsMap<{[P in keyof T]: Bool8}>)));
  }
```

modules/cudf/test/cudf-data-frame-tests.ts
```diff
@@ -437,3 +438,19 @@ test('dataframe.nansToNulls', () => {
 
   expect(result.get('b').nullCount).toEqual(1);
 });
+
+test('dataframe.isNull', () => {
+  const a      = Series.new({type: new Int32, data: [0, null, 2, 3, null]});
+  const b      = Series.new({type: new Float32, data: new Float32Buffer([NaN, 0, 3, NaN, 5])});
+  const c      = StringSeries.new([null, null, 'foo', 'bar', '']);
```
trxcllnt, line 445: Why not use `Series.new()`?
```suggestion
  const c      = Series.new([null, null, 'foo', 'bar', '']);
```

modules/cudf/src/data_frame.ts
mzegar: yeah saw the `reduce` and `forEach` used in other methods in dataframe, assumed that floats needed to be handled differently so left it as is.

modules/cudf/src/data_frame.ts
mzegar: Spoke about this over slack, leaving as is.

trxcllnt (approved):

modules/cudf/src/data_frame.ts
```diff
@@ -876,12 +876,40 @@ export class DataFrame<T extends TypeMap = any> {
...
+   * //    a: [false, true, false],
+   * //    b: [false, false, true],
+   * // }
+   * ```
+   */
+  isNull(): DataFrame<{[P in keyof T]: Bool8}> {
+    return new DataFrame(new ColumnAccessor(this.names.reduce((cols, name) => {
+      cols[name] = this.get(name).isNull()._col;
+      return cols;
+    }, {} as ColumnsMap<{[P in keyof T]: Bool8}>)));
```
trxcllnt, line 913: I liked the style in https://github.com/rapidsai/node-rapids/pull/189. Does it work here too?

```suggestion
    return new DataFrame(this.names.reduce((cols, name) => ({
      ...cols,
      [name]: this.get(name).isNull()
    }), {} as SeriesMap<{[P in keyof T]: Bool8}>));
```

mzegar: Yeah, a bit neater.

PR #193, opened 2021-05-13
[FEA] `DataFrame` Math Ops: Trig, Inverse, and Hyperbolic

mzegar:
**Changes**
- Adds trig, inverse, and hyperbolic functions to `dataframe`

**Unknown**
- Do we want to pass in a `memoryResource` to all these?

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -858,12 +858,124 @@ export class DataFrame<T extends TypeMap = any> {
   }
 
   /**
+   * Compute the trigonometric sine for all NumericSeries in the DataFrame
+   *
+   * @returns A DataFrame with the operation performed on all NumericSeries
+   * @example
+   * ```typescript
+   * import {DataFrame, Series, Int8}  from '@rapidsai/cudf';
+   *
+   * const df = new DataFrame({
+   *  a: Series.new({type: new Int8, data: [-3, 0, 3]});
+   * });
+   * df.sin();
+   * // return {
+   * //    a: [0, 0, 0],
+   * // }
+   * ```
+   */
+  sin(): DataFrame<T> {
+    const mapper = (map: SeriesMap<T>, name: string) => {
+      return {
+        ...map,
+        [name]:
+          this._performMathOp(name, () => { return Series.new(this._accessor.get(name).sin()); })
+      };
+    };
+    return new DataFrame(this.names.reduce(mapper, {} as SeriesMap<T>));
+  }
+
+  /**
+   * Compute the trigonometric cosine for all NumericSeries in the DataFrame
+   *
+   * @returns A DataFrame with the operation performed on all NumericSeries
+   * @example
+   * ```typescript
+   * import {DataFrame, Series, Int8}  from '@rapidsai/cudf';
+   *
+   * const df = new DataFrame({
+   *  a: Series.new({type: new Int8, data: [-3, 0, 3]});
+   * });
+   * df.cos();
+   * // return {
+   * //    a: [0, 1, 0],
+   * // }
+   * ```
+   */
+  cos(): DataFrame<T> {
+    const mapper = (map: SeriesMap<T>, name: string) => {
+      return {
+        ...map,
+        [name]:
+          this._performMathOp(name, () => { return Series.new(this._accessor.get(name).cos()); })
+      };
+    };
+    return new DataFrame(this.names.reduce(mapper, {} as SeriesMap<T>));
+  }
+
+  /**
+   * Compute the trigonometric tangent for all NumericSeries in the DataFrame
+   *
+   * @returns A DataFrame with the operation performed on all NumericSeries
+   * @example
+   * ```typescript
+   * import {DataFrame, Series, Int8}  from '@rapidsai/cudf';
+   *
+   * const df = new DataFrame({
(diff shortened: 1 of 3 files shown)


modules/cudf/test/cudf-data-frame-tests.ts
```diff
@@ -426,6 +427,20 @@ test(
...
+  const a  = Series.new({type: new Int32, data: [-3, 0, 3]});
+  const b  = Series.new({type: new Int8, data: [-3, 0, 3]});
+  const c  = Series.new(['foo', null, 'bar']);
+  const df = new DataFrame({'a': a, 'b': b, 'c': c});
+
+  test('dataframe.acos', () => {
+    const result = df.acos();
+    expect([...result.get('a')]).toEqual([-2147483648, 1, -2147483648]);
+    expect([...result.get('b')]).toEqual([0, 1, 0]);
+    expect([...result.get('c')]).toEqual([...c]);
```
bryevdv, line 440: Could add a floating point series, confirming that the value in the returned result dataframe is the same as calling `acos` on the series directly:
```
const d  = Series.new({type: new Float32, data: [-2.7, 0, 3.1]});

expect([...result.get('d')]).toEqual([...d.acos()]);
```
This is not as nice as a direct test of values but it does least ensure that the DataFrame and Series calculations on floating point are consistent with each other

modules/cudf/src/data_frame.ts
```diff
@@ -847,6 +848,33 @@ export class DataFrame<T extends TypeMap = any> {
...
+   */
+  acos(): DataFrame<T> {
+    return new DataFrame(this.names.reduce((map, name) => ({
+                                             ...map,
+                                             [name]: this.get(name) instanceof NumericSeries
+                                                       ? Series.new(this._accessor.get(name).acos())
+                                                       : Series.new(this._accessor.get(name))
+                                           }),
+                                           {} as SeriesMap<T>));
+  }
```
bryevdv, line 876: This is lovely. I might personally find it a little more friendly to read more like this:
```typescript
  acos(): DataFrame<T> {
    const mapper = (map: SeriesMap<T>, name: string) => {
      return {
        ...map, [name]: this.get(name) instanceof NumericSeries
                  ? Series.new(this._accessor.get(name).acos())
                  : Series.new(this._accessor.get(name))
      }
    };
    return new DataFrame(this.names.reduce(mapper, {} as SeriesMap<T>));
  }
```

Which also makes me wonder if maybe there could even be a factory function to generate the mapper, so that all the methods could just be something like
```typescript
 return new DataFrame(this.names.reduce(this._make_mapper("acos"), {} as SeriesMap<T>));
```
but @trxcllnt may think that is too clever by half :grinning:

mzegar: Yeah I'm open to whether or not this should be condensed even more. I was considering it as something similar is done in the tests themselves...

https://github.com/rapidsai/node-rapids/blob/781d9f12465ce0d570e14e91cf5f67794b430f3b/modules/cudf/test/series/unaryop/int8-tests.ts#L37

bryevdv: >  I'm assuming the only way to implement such a function would be to pass in a string of the method name

AFAIK, which would definitely be the biggest drawback. But 10-100s of lines of boilerplate has its own drawbacks as well, so it's definitely a trade-off. @trxcllnt might have some clever way to make the typing situation better, though.

mzegar: Might have been overthinking this, I pushed some new examples of how `acos` could potentially look. I think the current definition of `acos()` is probably best. (Unless there is something else I can abstract away)

modules/cudf/src/data_frame.ts
```diff
@@ -847,6 +848,63 @@ export class DataFrame<T extends TypeMap = any> {
     }
   }
 
+  /**
+   * Perform a math operation on a column that is of type NumericSeries
+   * @ignore
+   */
+  _performMathOp<R extends DataType>(columnName: string, mathOp: () => Series<R>): Series<R> {
```
trxcllnt, line 855: seems like `columnName: string` here is widening the type from `keyof T` to any string?

modules/cudf/src/data_frame.ts
```diff
@@ -847,6 +848,63 @@ export class DataFrame<T extends TypeMap = any> {
...
+    if (this.get(columnName) instanceof NumericSeries) { return mathOp(); }
+    return Series.new(this._accessor.get(columnName));
+  }
```
trxcllnt, line 858: Could this be a standalone local utility function instead of a public method on `DataFrame`?

modules/cudf/src/data_frame.ts
trxcllnt: Could this be a standalone local utility function instead of a public method on `DataFrame`?

trxcllnt (approved):

trxcllnt:
> @mzegar Do we want to pass in a memoryResource to all these?

Yeah we really should.

trxcllnt (changes requested):
We should allow passing a `memoryResource` into the functions.

mzegar:
I'm going to push these changes since I don't want to deal with any merge conflicts on other branch. Let me know if anything needs to be changed, but adding in memory resource shouldn't have caused any weird changes.

PR #197, opened 2021-05-14
[FIX] docker and dev env quality

trxcllnt:
* Set `CCACHE_CONFIGPATH` env var so `ccache -s` works in the container
* Set `DOCKER_WORKDIR` env var so shell history in VSCode Remote Container terminals is preserved
* Allow building the devel container with different LLDB, clangd, and clang-format versions (for testing/upgrading)
* Show the `.cmake-js` headers folder in the file explorer
* Don't launch 4 clangd processes per file
* Auto-detect VSCode workspace tasks

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,6 +59,9 @@ services:
         CMAKE_VERSION: "${CMAKE_VERSION:-3.20.2}"
         CCACHE_VERSION: "${CCACHE_VERSION:-4.1}"
         BASE_IMAGE: nvidia/cudagl:${CUDA_VERSION:-11.0}-devel-${LINUX_VERSION:-ubuntu18.04}
+    environment:
+      <<: *common_environment_variables
+      DOCKER_WORKDIR: "${PWD}"
     volumes:
       - *workdir
       - *etc_fonts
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -58,6 +59,9 @@ services:
         CMAKE_VERSION: "${CMAKE_VERSION:-3.20.2}"
         CCACHE_VERSION: "${CCACHE_VERSION:-4.1}"
         BASE_IMAGE: nvidia/cudagl:${CUDA_VERSION:-11.0}-devel-${LINUX_VERSION:-ubuntu18.04}
+    environment:
+      <<: *common_environment_variables
+      DOCKER_WORKDIR: "${PWD}"
```
ajschmidt8, line 64: any reason you chose not to include `DOCKER_WORKDIR` in `common_environment_variables`?

bryevdv (approved):

ajschmidt8 (approved):

docker-compose.yml
trxcllnt: I didn't want to set it for the `runtime` containers

PR #198, opened 2021-05-14
[FIX] make GraphCOO `numNodes` and `numEdges` functions instead of getters

trxcllnt:
Fixes node REPL crashes reported in https://github.com/nodejs/node-addon-api/issues/1000.

### DIFF
--- a/modules/cugraph/src/graph_coo.cpp
+++ b/modules/cugraph/src/graph_coo.cpp
@@ -14,9 +14,12 @@
 
 #include <node_cugraph/graph_coo.hpp>
 
+#include <node_cudf/utilities/dtypes.hpp>
+
 #include <node_cuda/utilities/error.hpp>
 #include <node_cuda/utilities/napi_to_cpp.hpp>
 
+#include <cudf/reduction.hpp>
 #include <cudf/types.hpp>
 
 #include <napi.h>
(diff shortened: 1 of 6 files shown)


modules/cugraph/src/graph_coo.cpp
```diff
@@ -56,13 +67,10 @@ GraphCOO::GraphCOO(CallbackArgs const& args) : EnvLocalObjectWrap<GraphCOO>(args
...
-    auto const& src      = *src_.Value();
-    auto const& dst      = *dst_.Value();
-    auto const src_max   = src.minmax().second;
-    auto const dst_max   = dst.minmax().second;
-    node_count_          = 1 + std::max<int32_t>(  //
-                        src_max->get_value().ToNumber(),
-                        dst_max->get_value().ToNumber());
+    auto max_id = [&](Napi::Reference<Column::wrapper_t> const& col) -> int32_t {
+      return col.Value()->minmax().second->get_value().ToNumber();
+    };
```
bryevdv, line 72: Wasn't there some obscure problem related to lifetimes that precipitated the old step-by-step code structure? Does this new version with the lambda avoid that?

trxcllnt: No there isn't, but I refactored it into a lambda to verify that, and left it since it's a bit cleaner.

bryevdv (approved):

PR #206, opened 2021-05-25
Update Bindings Table

ajschmidt8:
This PR updates the binding table to check off `Series.unique` and `Series.value_counts`, which were done in #143.


mzegar:
Do these need to be marked off as well inside of `List Series`, `String Series`, and `Struct Series`?

Others LGTM.

mzegar (approved):

ajschmidt8:
> Do these need to be marked off as well inside of `List Series`, `String Series`, and `Struct Series`?
> 
> Otherwise LGTM.

good point! I will update those as well.

PR #208, opened 2021-05-25
[FEA] Add Categorical DataType + Series

trxcllnt:
Adds a Categorical `DataType` that uses `cudf::type_id::DICTIONARY32`. `DICTIONARY32` is a specialization of Arrow's Dictionary type, but where the keys must be an INT32 column.

Categorical columns allow the user to save memory by compacting duplicate values into a dictionary (or list of "categories"), and using a corresponding map of integer value positions to their dictionary positions:

```typescript
import { Series, Categorical } from '@rapidsai/cudf'

const strs  = Series.new(['foo', 'foo', 'bar', 'bar'])
const categorical = strs.cast(new Categorical(strs.type))

[...categorical] // ['foo', 'foo', 'bar', 'bar']
[...categorical.codes] // [1, 1, 0, 0]
[...categorical.categories] // ['bar', 'foo']
```

Other changes in this PR:
* Updates `DataFrame.join` options to make `how` an optional key, pass memoryResource down into C++ `Table.join()`.
* Update `OrderSpec` type to make `ascending` and `null_order` optional to match `Series.orderBy` signature, default `null_order` to `AFTER` (to match cuDF Python).
* Adds `Series.encodeLabels()` implementation to support encoding arbitrary types as Categoricals.

### DIFF
--- a/modules/cudf/src/series.ts
+++ b/modules/cudf/src/series.ts
@@ -24,6 +24,7 @@ import {Scalar} from './scalar';
 import {Table} from './table';
 import {
   Bool8,
+  Categorical,
   DataType,
   Float32,
   Float64,
@@ -32,6 +33,7 @@ import {
   Int32,
   Int64,
   Int8,
+  Integral,
   List,
   Numeric,
   Struct,
@@ -140,7 +142,7 @@ export type Series<T extends arrow.DataType = any> = {
   [arrow.Type.FixedSizeBinary]: never,  // TODO
   [arrow.Type.FixedSizeList]: never,    // TODO
   [arrow.Type.Map]: never,              // TODO
-  [arrow.Type.Dictionary]: never,       // TODO
+  [arrow.Type.Dictionary]: CategoricalSeries<(T extends arrow.Dictionary ? T['valueType'] : any)>
 }[T['TType']];
 
 /**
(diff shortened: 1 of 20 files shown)


modules/cudf/src/series/string.ts
```diff
@@ -25,15 +26,25 @@ export class StringSeries extends Series<Utf8String> {
...
+    if (arrow.DataType.isDictionary(type)) {
+      const vals = this.cast(type.dictionary).unique(true, memoryResource);
+      const keys = this.encodeLabels(vals, undefined, undefined, memoryResource);
+      return Series.new<R>(new Column({
+        type: new Categorical(type.dictionary) as R,
+        length: keys.length,
+        nullMask: this.mask,
+        children: [keys._col, vals._col]
+      }));
+    }
```
trxcllnt, line 45: Refactoring this casting logic is planned in a future PR. The idea is the top-level `Series` will have a `cast()` method, which will check the dtype and call an appropriate protected `_castTo<Type>()` function, which each Series specialization can override:

```typescript
class Series {
  public cast(dataType, memoryResource?) {
    switch (dataType.typeId) {
      case arrow.Type.Integer: return this._castToInt(dataType, memoryResource);
      // etc.
    }
  }
  protected castToInt() { throw new Error("unimplemented"); }
  // etc.
}

class StringSeries extends Series {
  protected castToInt(dataType, memoryResource) {
    return Series.new(this._col.stoi(dataType, memoryResource));
  }
}
```

modules/cudf/test/series/categorical-tests.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  test('Can cast a String Series to Categorical', () => {
+    const categories  = Series.new(['foo', 'foo', 'bar', 'bar']);
+    const categorical = categories.cast(new Categorical(categories.type));
+    expect([...categorical]).toEqual(['foo', 'foo', 'bar', 'bar']);
+    expect([...categorical.codes]).toEqual([1, 1, 0, 0]);
+    expect([...categorical.categories]).toEqual(['bar', 'foo']);
+  });
+});
```
bryevdv, line 47: I think a test for `setCategories` is advised, given its complexity

modules/cudf/src/series.ts
```diff
@@ -399,6 +402,58 @@ export class AbstractSeries<T extends DataType = any> {
...
+   * Encode the Series values into integer labels.
+   *
+   *
+   * @param categories Th optional Series of values to encode into integers. Defaults to the unique
+   *   elements in this Series.
+   * @param type The optional integer DataType to use for the returned Series. Defaults to Int32.
+   * @param nullSentinel The optional value used to indicate missing category. Defaults to -1.
+   * @param memoryResource The optional MemoryResource used to allocate the result Column's device
+   *   memory.
+   * @returns A sequence of encoded integer labels with values between `0` and `n-1` categories
```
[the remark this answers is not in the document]
trxcllnt: Oh yeah, it definitely can. I think I copied this docstring from Python :sweat_smile:

modules/cudf/test/series/categorical-tests.ts
trxcllnt: `setCategories` isn't really ready yet. It's one part of the larger feature of "messing with categories," and I was going to add those in a follow-up PR too. I originally started doing it, decided it was too much work to stuff into this PR, but left it in because I didn't want to rewrite the code.

modules/cudf/src/series.ts
```diff
@@ -399,6 +402,58 @@ export class AbstractSeries<T extends DataType = any> {
    */
   countNonNulls(): number { return this._col.length - this._col.nullCount; }
 
+  /**
+   * Encode the Series values into integer labels.
+   *
+   *
+   * @param categories Th optional Series of values to encode into integers. Defaults to the unique
```
mzegar, line 409: `Th` -> `The`

bryevdv (approved):

AjayThorve (approved):
LGTM

PR #212, opened 2021-05-26
[FEA] Dataframe `concat`

mzegar:
Closes https://github.com/rapidsai/node-rapids/issues/97

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -70,7 +70,7 @@ function _invokeIfNumericSeries<P extends keyof T, T extends TypeMap, R extends 
   return Series.new(series._col as Column<R>);
 }
 
-function concat<DFs extends DataFrame[]>(...dfs: DFs) {
+function concat<DFs extends DataFrame[], T extends TypeMap>(...dfs: DFs): DataFrame<T> {
   const allNames = [...dfs
                       .reduce((namesMap, df) => df.names.reduce(
                                 (namesMap, name) => namesMap.set(name, true), namesMap),
@@ -138,7 +138,10 @@ function concat<DFs extends DataFrame[]>(...dfs: DFs) {
   const tables: Table[] = [];
   columnsPerDF.forEach((column) => { tables.push(new Table({columns: column})); });
   const concatenatedTable = Table.concat(tables);
-  console.log(concatenatedTable);
+
+  return new DataFrame(allNames.reduce(
+    (map, name, index) => ({...map, [name]: Series.new(concatenatedTable.getColumnByIndex(index))}),
+    {} as SeriesMap<T>));
 }
 
 /**
@@ -417,7 +420,7 @@ export class DataFrame<T extends TypeMap = any> {
   }
 
   // TODO: MZEGAR write docstrings
-  concat<DFs extends DataFrame[]>(...dfs: DFs) { return concat(...[this, ...dfs]); }
+  concat<DFs extends DataFrame[]>(...dfs: DFs): DataFrame<T> { return concat(...[this, ...dfs]); }
 
   /**
    * Generate an ordering that sorts DataFrame columns in a specified way
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -1711,7 +1711,7 @@ export class DataFrame<T extends TypeMap = any> {
   dropDuplicates(keep: keyof typeof DuplicateKeepOption,
                  nullsEqual: boolean,
                  nullsFirst: boolean,
-                 subset: (string&keyof T)[] = this.names,
+                 subset = this.names,
                  memoryResource?: MemoryResource) {
     const column_indices: number[] = [];
     const allNames                 = this.names;
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -116,7 +116,7 @@ function _concat<DFs extends DataFrame[], T extends TypeMap>(...dfs: DFs): DataF
    * [Float64, Int32, Float64]
    * ```
    */
-  const common_dtypes: any[] = first_non_null_dtype.map((tuple) => { return tuple[1]; });
+  const common_dtypes: DataType[] = first_non_null_dtype.map((tuple) => { return tuple[1]; });
   first_non_null_dtype.forEach((tuple, col_idx) => {
     const first_non_null_dtype_idx = tuple[0];
     const start_idx                = first_non_null_dtype_idx + 1;
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -70,12 +70,12 @@ function _invokeIfNumericSeries<P extends keyof T, T extends TypeMap, R extends 
   return Series.new(series._col as Column<R>);
 }
 
-function concat<DFs extends DataFrame[], T extends TypeMap>(...dfs: DFs): DataFrame<T> {
-  const allNames = [...dfs
-                      .reduce((namesMap, df) => df.names.reduce(
-                                (namesMap, name) => namesMap.set(name, true), namesMap),
-                              new Map<string, boolean>())
-                      .keys()];
+function concat<DFs extends DataFrame[], T extends TypeMap>(...dfs: DFs): DataFrame {
+  const all_column_names = [...dfs
+                              .reduce((namesMap, df) => df.names.reduce(
+                                        (namesMap, name) => namesMap.set(name, true), namesMap),
+                                      new Map<string, boolean>())
+                              .keys()];
 
   /**
    * Array<Array<(Column|null)>> -- If a DF has a Column for a given name, it will be in the list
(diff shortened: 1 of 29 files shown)


mzegar:
Attempting to pass in an array

``ts
const result = df.concat([df2]);
```

Throws the error 
```
Argument of type 'DataFrame<{ a: Float64; }>[]' is not assignable to parameter of type 'DataFrame<any>'.
  Type 'DataFrame<{ a: Float64; }>[]' is missing the following properties from type 'DataFrame<any>': _accessor, numRows, numColumns, names, and 45 more.
 ```

modules/cudf/src/data_frame.ts
```diff
@@ -70,6 +70,93 @@ function _invokeIfNumericSeries<P extends keyof T, T extends TypeMap, R extends
...
+   *
+   * columnsPerDF == [
+   *  [dfs[0].get("a"), dfs[0].get("b"),            null],
+   *  [           null, dfs[1].get("b"), dfs[1].get("c")]
+   * ]
+   * ```
+   */
+  const columns_per_df: any[][] =
+    dfs.map((df) => all_column_names.map((name) => df.has(name) ? df.get(name)._col : null));
+
```
mzegar, line 94: Throw here if `columns_per_df` is empty?

mzegar, line 94: What do we want to do if `columns_per_df` is empty? Throw? return early?

modules/cudf/src/data_frame.ts
```diff
@@ -70,6 +70,93 @@ function _invokeIfNumericSeries<P extends keyof T, T extends TypeMap, R extends
...
+   * ```
+   */
+  const common_dtypes: DataType[] = first_non_null_dtype.map((tuple) => { return tuple[1]; });
+  first_non_null_dtype.forEach((tuple, col_idx) => {
+    const first_non_null_dtype_idx = tuple[0];
+    const start_idx                = first_non_null_dtype_idx + 1;
+    for (let row_idx = start_idx; row_idx < num_of_rows; ++row_idx) {
+      const first_non_null_dtype = tuple[1];
+      const col                  = columns_per_df[row_idx][col_idx];
+      if (col !== null) { common_dtypes[col_idx] = findCommonType(first_non_null_dtype, col.type); }
```
mzegar, line 126: Assuming if `first_non_null_dtype` has nulls, `findCommonType` should throw if passed in a null.

modules/cudf/src/data_frame.ts
```diff
@@ -70,6 +70,93 @@ function _invokeIfNumericSeries<P extends keyof T, T extends TypeMap, R extends
...
+  /**
+   * If any columns are null, create an empty column with type of common dtype
+   * Otherwise cast the column using the common dtype (if it isn't already the correct type).
+   */
+  common_dtypes.forEach((common_dtype, col_idx) => {
+    for (let row_idx = 0; row_idx < num_of_rows; ++row_idx) {
+      const col = columns_per_df[row_idx][col_idx];
+      if (col === null) {
+        const empty_column_length =
+          columns_per_df.find((row) => row !== null)?.find((col) => col !== null).length;
```
mzegar, line 139: Could be the case that `find` returns a []. Creating an array of size `0`, and then throwing once the tables are being constructed. Not sure if we want to handle specifically with a throw.

modules/cudf/src/data_frame.ts
```diff
@@ -70,6 +70,93 @@ function _invokeIfNumericSeries<P extends keyof T, T extends TypeMap, R extends
...
+          columns_per_df[row_idx][col_idx] = col.cast(common_dtype);
+        }
+      }
+    }
+  });
+
+  const tables: Table[] = [];
+  columns_per_df.forEach((col) => { tables.push(new Table({columns: col})); });
+
+  const concatenatedTable = Table.concat(tables);
```
mzegar, line 153: Got any cool `.js` syntactic sugar to condense this?

modules/cudf/src/table/concatenate.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <cudf/concatenate.hpp>
+#include <node_cudf/table.hpp>
+
+namespace nv {
+
+Napi::Value Table::concat(Napi::CallbackInfo const& info) {
+  CallbackArgs args{info};
+
+  std::vector<Table::wrapper_t> tables = args[0];
+  rmm::mr::device_memory_resource* mr  = args[1];
```
mzegar, line 24: Do we want to check the args using `NODE_CUDA_EXPECT`? I don't think it would be possible to call it with anything else.

mzegar (commented):
Got some general questions

modules/cudf/src/data_frame.ts
bryevdv: I haven't checked cudf but pandas raises an exception in this case.

modules/cudf/src/table/concatenate.cpp
bryevdv: Maybe check that `tables` is at least length 1?

modules/cudf/src/data_frame.ts
```diff
@@ -345,6 +432,31 @@ export class DataFrame<T extends TypeMap = any> {
...
+   * });
+   *
+   * const df2 = new DataFrame({
+   *   b: Series.new([5, 6, 7, 8]),
+   * });
+   *
+   * df.concat(df2);
+   * // return {
+   * //    a: [1, 2, 3, 4, null, null, null, null],
+   * //    b: [null, null, null, null, 5, 6, 7, 8],
```
bryevdv, line 454: Oh, hrm , this is actually the case I meant with the comment about pandas above. It you try to concat tow df's that have no columns in common at all, it raises an error. Maybe we do want to do this for node-rapids, though

trxcllnt: Reminder we shouldn't target Pandas behavior, except incidentally due to targeting cuDF Python.

modules/cudf/src/data_frame.ts
trxcllnt, line 153:
```suggestion
  const concatenatedTable = Table.concat(columns_per_df.map((columns) => new Table({ columns })));
```

mzegar: Thanks, this is what I was looking for.

modules/cudf/src/data_frame.ts
mzegar: so I'm assuming keeping it the same is fine then, this is how `cuDF` handles it when attempting it on my end.

bryevdv: Definitely, I compared to pandas because I happened to be on my laptop at the time, but I should have waited to comment until I could have checked cudf on the workstation

modules/cudf/src/table/concatenate.cpp
mzegar: Sorry, just wanted to clarify, throw a `NODE_CUDA_EXPECT` if `tables.size() == 0`?

bryevdv: Yes that's what I meant to propose. It shouldn't ever happen given the construction of `DataFame.concat` supplying `this` but it seems like a good precaution anyway (cc @trxcllnt)

modules/cudf/src/table/concatenate.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+  std::vector<cudf::table_view> table_views;
+  table_views.reserve(tables.size());
+  std::transform(tables.begin(),
+                 tables.end(),
+                 std::back_inserter(table_views),
+                 [](const Table::wrapper_t& t) -> cudf::table_view { return t->view(); });
+
+  return Table::New(info.Env(), cudf::concatenate(table_views, mr))->Value();
```
bryevdv, line 33: I think this needs a `try...catch` since `cudf::concatenate` can potentially throw a cudf::logic_error:

https://docs.rapids.ai/api/libcudf/stable/group__copy__concatenate.html#ga172353e7673decae3d2ebad09e003832

mzegar: So apparently the creation of a table with mismatch columns also does this check.

`NODE_CUDF_EXPECT(col->size() == num_rows_, "All Columns must be of same length");`
in `table.cpp` `Table::Table(...)`

So that ends up throwing before it even gets to my `concat` implementation.

bryevdv: isn't the error condition if the table views have differing number columns (not about the length columns)?

mzegar: spoke with Bryan and he cleared up my misunderstandings...

1. Using the Table constructor and providing columns of mismatched sizes leads to CUDF blowing up. I'll handle this in the PR with a try catch
2. The public method `concat` inside of `Table.ts` can be called with an array of tables of potentially mismatched sizes. We should also wrap this in a try .. catch.

A side conversation emerged of whether or not we want to allow users to access `Table.ts` & `Column.ts`, I think we have an issue posted about this.

modules/cudf/src/table/concatenate.cpp
mzegar: Spoke with Bryan on Friday about the difference between...

```
NODE_CUDF_EXPECT(col->size() == num_rows_, "All Columns must be of same length");

vs
.
NAPI_THROW(Napi::Error::New(args.Env(),
    "Table constructor requires all columns to be of same length"));
```

Is there any reason why it seems they throw in two different ways? I can setup an example if needed, but these were the outputs...

```
NODE_CUDF_EXPECT(col->size() == num_rows_, "All Columns must be of same length");

terminate called after throwing an instance of 'cudf::logic_error'
  what():  cuDF failure:
All Columns must be of same length
    at ../../src/table.cpp:89
Aborted (core dumped)
```

```
NAPI_THROW(Napi::Error::New(args.Env(),
    "Table constructor requires all columns to be of same length"));

    Table constructor requires all columns to be of same length
      753 |     const b = Series.new([5, 6, 7, 8, 9]);
      754 |
    > 755 |     const result = new Table({columns: [a._col, b._col]});
          |                    ^
      756 |     console.log(result);
      757 |   });
      758 | });
      at Object.<anonymous> (test/cudf-data-frame-tests.ts:755:20)
```

modules/cudf/src/table.cpp
```diff
@@ -86,7 +86,10 @@ Table::Table(CallbackArgs const& args) : EnvLocalObjectWrap<Table>(args) {
     num_rows_ = col->size();
     for (auto i = 1u; i < columns.Length(); ++i) {
       col = Column::wrapper_t{columns.Get(i).ToObject()};
-      NODE_CUDF_EXPECT(col->size() == num_rows_, "All Columns must be of same length");
+      if (!(col->size() == num_rows_)) {
+        NAPI_THROW(Napi::Error::New(args.Env(),
+                                    "Table constructor requires all columns to be of same length"));
+      }
```
trxcllnt, line 92:
```suggestion
      NODE_CUDF_EXPECT(col->size() == num_rows_, "All Columns must be of same length", args.Env());
```

modules/cudf/src/table/concatenate.cpp
mzegar: Solved, `args.Env()` needs to be passed in.

`NODE_CUDF_EXPECT(col->size() == num_rows_, "All Columns must be of same length", args.Env());`

modules/cudf/src/data_frame.ts
```diff
@@ -70,6 +70,89 @@ function _invokeIfNumericSeries<P extends keyof T, T extends TypeMap, R extends
   return Series.new(series._col as Column<R>);
 }
 
+function _concat<DFs extends DataFrame[], T extends TypeMap>(...dfs: DFs): DataFrame {
```
trxcllnt, line 73: We should be able to infer `T` from the combined names in the `DFs` type. If not, we might need to rethink this generic type signature.

```suggestion
function _concat<DFs extends DataFrame[]>(...dfs: DFs): DataFrame {
```

mzegar: Hmm... spent some time with this, but the complexity with typescript types still eludes me. Able to give a little more info?

I'm assuming this is what we want
1. Have some way to keep the types of each of the columns (for every single dataframe passed in)
2. Combine these types together at the end using `&` (similar to join).
3. Typescript should show how many columns result in the concat, and their types.

Assuming what I've said is even possible.

bryevdv:
Is the empty ` libnccl.so.2` intentional?

mzegar:
> Is the empty ` libnccl.so.2` intentional?

Nope, forgot to delete this, good catch, thanks.

modules/cudf/src/dataframe/concat.ts
```diff
@@ -0,0 +1,115 @@
...
+// clang-format on
+
+export function concat<TFirst extends DataFrame, TRest extends DataFrame[]>(first: TFirst,
+                                                                            ...others: TRest) {
+  const dfs              = [first, ...others] as [TFirst, ...TRest];
+  const all_column_names = [...dfs
+                              .reduce((namesMap, df) => df.names.reduce(
+                                        (namesMap, name) => namesMap.set(name, true), namesMap),
+                                      new Map<string, boolean>())
+                              .keys()];
```
bryevdv, line 36: I think maybe this could be simplified?
```
const all_column_names = [...new Set(dfs.reduce((names, df) => names.concat(df.names), [])).keys()]
```

modules/webgl/src/webgl.ts
```diff
@@ -144,10 +144,10 @@ function getShaderInfoLog(this: WebGL2RenderingContext, shader: WebGLShader): st
       let errIndex, numIndex, errMatch, numMatch, type, num;
       ({index: errIndex, [0]: errMatch, [1]: type} =
          // eslint-disable-next-line @typescript-eslint/prefer-regexp-exec
-       (lines[0] || '').match(/\s?(warning|error) (\w+|\d+):/) || {...['', '']});
+       (lines[0] || '').match(/\s?(warning|error) (\w+|\d+):/) || {index: undefined, 0: '', 1: ''});
       ({index: numIndex, [0]: numMatch, [1]: num} =
          // eslint-disable-next-line @typescript-eslint/prefer-regexp-exec
-       (line.match(/0\((\d+)\) :/) || {...['', '']}));
+       (line.match(/0\((\d+)\) :/) || {index: undefined, 0: '', 1: ''}));
```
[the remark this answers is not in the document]
mzegar: @trxcllnt updated typescript in this branch, had to fix some new type errors that came along with the new versioning.

modules/cudf/src/dataframe/concat.ts
mzegar:
```
No overload matches this call.
  Overload 1 of 2, '(...items: ConcatArray<never>[]): never[]', gave the following error.
    Argument of type 'readonly string[]' is not assignable to parameter of type 'ConcatArray<never>'.
      The types returned by 'slice(...)' are incompatible between these types.
        Type 'string[]' is not assignable to type 'never[]'.
          Type 'string' is not assignable to type 'never'.
  Overload 2 of 2, '(...items: ConcatArray<never>[]): never[]', gave the following error.
    Argument of type 'readonly string[]' is not assignable to parameter of type 'ConcatArray<never>'.ts(2769)
 ```
 
 Over `(df.names)`
 
 As parameter `names` ends up being `never[]`. Unless I'm misunderstanding, isn't `dfs` just an array of `DataFrame`?

bryevdv: > isn't dfs just an array of DataFrame

I thought so. FWIW I was just basing my suggestion on some toy experiments:

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDA3mOBbAphAXDAbQCIAzEEIgGhiICM4AnIgXQF8AoUSWWpVdbHkJ1GVGvQBeLDp3DQYAExIQ+BONVrN2s7jACuYAJYBHPVgD6AnKoB0dsFgDuMAMpYoACiUQbDLAr1gLA8PKwhqJQBKJAA+GDCbLmA4TyUbMMjqAmZIyJsAaywATwgPSK0uCBAAGywbapAAcw8DEzNLTBxIoA

So it's possible (well, definitely) missed some detail

mzegar: Spoke with Bryan, found a stackoverflow thread that ran into this too, just define `names` as type `string[]`.

bryevdv (approved):

trxcllnt (approved):
Looks good to me. We should split out and add more concat tests in a follow-up PR.

PR #213, opened 2021-05-27
Series/DataFrame `astype` `cast`

mzegar:
`cast` already exists, write tests and update the bindings doc.


ajschmidt8 (approved):
Tests look good and passed when I ran them locally :+1:

PR #215, opened 2021-05-31
[FEA] Added dataframe-replaceNulls

mzegar:
**Changes**
- Added dataframe `replaceNulls`
- Added dataframe `replaceNulls` test

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -1610,7 +1610,7 @@ export class DataFrame<T extends TypeMap = any> {
    * // }
    * ```
    */
-  replaceNulls(value: number|string|boolean, memoryResource?: MemoryResource): DataFrame;
+  replaceNulls(value: number|string|boolean, memoryResource?: MemoryResource): DataFrame<T>;
 
   /**
    * Replace null values with the corresponding elements from another Map of Series.
@@ -1635,7 +1635,7 @@ export class DataFrame<T extends TypeMap = any> {
    * // }
    * ```
    */
-  replaceNulls(value: SeriesMap<T>, memoryResource?: MemoryResource): DataFrame;
+  replaceNulls(value: SeriesMap<T>, memoryResource?: MemoryResource): DataFrame<T>;
 
   /**
    * Replace null values with a scalar value.
@@ -1660,7 +1660,7 @@ export class DataFrame<T extends TypeMap = any> {
    * // }
    * ```
    */
-  replaceNulls(value: any, memoryResource?: MemoryResource): DataFrame {
+  replaceNulls(value: any, memoryResource?: MemoryResource): DataFrame<T> {
     if (value instanceof Object) {
       const columns = new ColumnAccessor(_seriesToColumns(value as SeriesMap<T>));
       return new DataFrame(this.names.reduce(
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -1748,9 +1748,10 @@ export class DataFrame<T extends TypeMap = any> {
    */
   replaceNulls(value: SeriesMap<T>, memoryResource?: MemoryResource): DataFrame<T>;
 
-  replaceNulls(value: any, memoryResource?: MemoryResource): DataFrame<T> {
+  replaceNulls<R extends DataType>(value: SeriesMap<T>|R['scalarType'],
+                                   memoryResource?: MemoryResource): DataFrame<T> {
     if (value instanceof Object) {
-      const columns = new ColumnAccessor(_seriesToColumns(value as SeriesMap<T>));
+      const columns = new ColumnAccessor(_seriesToColumns(value));
       return new DataFrame(this.names.reduce(
         (map, name) => ({
           ...map,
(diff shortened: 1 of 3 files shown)


modules/cudf/src/data_frame.ts
```diff
@@ -1587,4 +1587,75 @@ export class DataFrame<T extends TypeMap = any> {
...
+   * // return {
+   * //    a: [0, 1, 2],
+   * //    b: [1, 1, 1],
+   * // }
+   * ```
+   */
+  replaceNulls(value: SeriesMap<T>, memoryResource?: MemoryResource): DataFrame;
+
+  replaceNulls(value: any, memoryResource?: MemoryResource): DataFrame {
+    if (value instanceof Object) {
```
mzegar, line 1642: since `SeriesMap` is an interface, I'm unable to do a check against `instanceof`, any other suggestions?

modules/cudf/src/data_frame.ts
```diff
@@ -203,11 +204,11 @@ export class DataFrame<T extends TypeMap = any> {
    *
    * @example
    * ```typescript
-   * import {DataFrame} from '@rapidsai/cudf';
+   * import {DataFrame, Series} from '@rapidsai/cudf';
    *
    * const df = new DataFrame({a: [1, 2, 3]});
    *
-   * df.assign({b: ["foo", "bar", "bar"]})
+   * df.assign({b: Series.new(["foo", "bar", "bar"])})
```
mzegar, line 211: Assign uses a `SeriesMap`, which doesn't allow for a syntax such as...
```
df.assign({b: ["foo", "bar", "bar"]})
```
Maybe we should consider adding this.

trxcllnt: Good catch, this docstring was wrong.

modules/cudf/src/data_frame.ts
```diff
@@ -1587,4 +1588,76 @@ export class DataFrame<T extends TypeMap = any> {
...
+   * });
+   *
+   * df.replaceNulls(1);
+   * // return {
+   * //    a: [0, 1, 2],
+   * //    b: [1, 1, 1],
+   * // }
+   * ```
+   */
+  replaceNulls(value: T['scalarType']|any, memoryResource?: MemoryResource): DataFrame;
```
trxcllnt, line 1615: The TypeMap T doesn't have a `scalarType`, so this overload is invalid.

modules/cudf/src/data_frame.ts
```diff
@@ -1587,4 +1588,76 @@ export class DataFrame<T extends TypeMap = any> {
...
+   * df.replaceNulls({'a': Series.new([0, 1, 2]), 'b': Series.new([1, 1, 1])});
+   * // return {
+   * //    a: [0, 1, 2],
+   * //    b: [1, 1, 1],
+   * // }
+   * ```
+   */
+  replaceNulls(value: SeriesMap<T>, memoryResource?: MemoryResource): DataFrame;
+
+  replaceNulls(value: any, memoryResource?: MemoryResource): DataFrame {
```
trxcllnt, line 1642: Need to preserve the full generic type here. Otherwise consumers will get a type-erased DataFrame if they call this.

modules/cudf/src/data_frame.ts
mzegar: Any suggestions on replacements? is `any` fine?

or would `number|string|boolean` cover all the scalar types.

mzegar: Nevermind... I think our chats about datatypes helped out.

```
replaceNulls(value: (DataType&keyof T)['scalarType'],
               memoryResource?: MemoryResource): DataFrame<T>;
```

Since the key of T is a DataType, we should be able to just access the `scalarType`.

modules/cudf/src/data_frame.ts
bryevdv: I'm not sure about the type check, but I think the signature can be improved?
```
  replaceNulls<R extends DataType>(value: R['scalarType']|SeriesMap<T>,
                                   memoryResource?: MemoryResource): DataFrame<T> {
```

For checking against interfaces folks seem to variously suggest adding s type tag field to the object but I don't think we can do that safely since the names in the map can be arbitrary and might conflict with anything we chose. Maybe there is some way to check or the converse, i.e. is a scalar type? Not sure. I do think with the narrower signature the check against object is a little less worrisome, with the possible exception of struct columns? Their scalar type is an object too `scalarType: {[P in keyof T]: T[P]['scalarType']};`

mzegar: I'll see if the converse can be done,

Also sorry this comment is referencing an older version of this PR, this is how it currently looks...
```
replaceNulls(value: T['scalarType']|any, memoryResource?: MemoryResource): DataFrame;
```

Should be basically the same as you've suggested above (I'll double check this).

bryevdv: ah yes, sorry for not checking the current diff!

mzegar: Hmm, no luck. There isn't really a neat way to check against interfaces like you've suggested.

We can type guard the generic `any` method though...

```
replaceNulls<R extends DataType>(value: SeriesMap<T>|R['scalarType'],
                                 memoryResource?: MemoryResource): DataFrame<T> {
```

bryevdv (approved):

PR #216, opened 2021-05-31
[FEA] Added bindings for `head` and `tail`

mzegar:
Using `gather` seemed like the logical choice to me.

**Changes**
- Added series `head` & `tail`
- Added dataframe `head` & `tail`
- Added tests for both

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -24,7 +24,7 @@ import {AbstractSeries, Series} from './series';
 import {NumericSeries} from './series/numeric';
 import {Table, ToArrowMetadata} from './table';
 import {CSVToCUDFType, CSVTypeMap, ReadCSVOptions, WriteCSVOptions} from './types/csv';
-import {Bool8, DataType, Float32, Float64, IndexType, Numeric} from './types/dtypes';
+import {Bool8, DataType, Float32, Float64, IndexType, Int32, Numeric} from './types/dtypes';
 import {NullOrder} from './types/enums';
 import {ColumnsMap, CommonType, TypeMap} from './types/mappings';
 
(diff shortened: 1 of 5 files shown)


modules/cudf/src/data_frame.ts
```diff
@@ -450,6 +450,71 @@ export class DataFrame<T extends TypeMap = any> {
...
+   * // {a: [0, 1, 2, 3, 4], b: [0.0, 1.0, 2.0, 3.0, 4.0]}
+   *
+   * b.head(1);
+   * // {a: [0], b: [0.0]}
+   *
+   * a.head(-1);
+   * // throws index out of bounds error
+   * ```
+   */
+  head(n = 5): DataFrame {
```
trxcllnt, line 477: Need to preserve the generic type signature:

```suggestion
  head(n = 5): DataFrame<T> {
```

modules/cudf/src/data_frame.ts
```diff
@@ -450,6 +450,71 @@ export class DataFrame<T extends TypeMap = any> {
...
+   * // {a: [2, 3, 4, 5, 6], b: [2.0, 3.0, 4.0, 5.0, 6.0]}
+   *
+   * b.tail(1);
+   * // {a: [6], b: [6.0]}
+   *
+   * a.tail(-1);
+   * // throws index out of bounds error
+   * ```
+   */
+  tail(n = 5): DataFrame {
```
trxcllnt, line 510: Here too:

```suggestion
  tail(n = 5): DataFrame<T> {
```

trxcllnt (changes requested):
Need to preserve generic types.

modules/cudf/src/series.ts
```diff
@@ -627,6 +627,59 @@ export class AbstractSeries<T extends DataType = any> {
...
+   * b.head(1); // ["foo"]
+   * a.head(-1); // throws index out of bounds error
+   * ```
+   */
+  head(n = 5): Series<T> {
+    if (n < 0) { throw new Error('Index provided is out of bounds'); }
+    const selection = Series.new({
+      type: new Int32,
+      data: Array.from({length: n < this._col.length ? n : this._col.length}, (_, i) => i)
+    });
```
bryevdv, line 652: I think [`Series.sequence`](https://github.com/rapidsai/node-rapids/blob/main/modules/cudf/src/series.ts#L785) could be used here (and a above)

modules/cudf/test/cudf-data-frame-tests.ts
```diff
@@ -233,6 +233,72 @@ test('DataFrame.gather (indices)', () => {
   expect([...rb]).toEqual([...expected_b]);
 });
 
+describe('Dataframe.head', () => {
+  const a  = Series.new([0, 1, 2, 3, 4, 5, 6]);
+  const b  = Series.new([0.0, 1.1, 2.2, 3.3, 4.4, 5.5, 6.6]);
+  const c  = Series.new(['foo', null, null, 'bar', null, null, 'foo']);
+  const df = new DataFrame({'a': a, 'b': b, 'c': c});
+
+  test('default index', () => {
```
bryevdv, line 242: extreme nit: "index" seems like the wrong name/concept here (and throughout), even just "n" would be slightly more clear I think

modules/cudf/src/series.ts
mzegar: Neat.

bryevdv (approved):

mzegar:
Merging, since it seems like all PR feedback is addressed, let me know if anything else needs to be changed.

PR #219, opened 2021-06-01
Pass in env() to NODE_CUDF_EXPECT

mzegar:


modules/cudf/src/column.cpp
```diff
@@ -360,14 +360,15 @@ void Column::set_null_mask(Napi::Value const& new_null_mask, cudf::size_type new
     if (new_null_count > 0) {
       NODE_CUDF_EXPECT(new_mask->size() >= cudf::bitmask_allocation_size_bytes(this->size()),
                        "Column with null values must be nullable, and the null mask "
-                       "buffer size should match the size of the column.");
+                       "buffer size should match the size of the column.",
+                       Env());
```
mzegar, line 364: Should this be `new_null_mask.Env()`? What's the difference betwee that and `Env()`

Is one associated to the value being passed in, and the other just to the method?

trxcllnt (approved):

modules/cudf/src/column.cpp
trxcllnt: Yeah, `new_null_mask.Env()` could be a different `Napi::Env` instance than this one. I think passing `this->Env()` is the correct one to use.

mzegar: Nice, thanks.

PR #220, opened 2021-06-02
[FEA] Added dataframe `sum`

mzegar:
Finally ready to go :+1:


mzegar:
How does `skipna` even work? https://docs.rapids.ai/api/cudf/nightly/api.html#cudf.core.dataframe.DataFrame.sum

I attempted this example in my Jupyter notebook.

```python
df = cudf.DataFrame({'a': [5, None, 5]})
print(df.sum()) # 10

df2 = cudf.DataFrame({'a': [5, None, 5]})
print(df2.sum(False)) # 10
```

modules/cudf/src/data_frame.ts
```diff
@@ -1494,6 +1494,22 @@ export class DataFrame<T extends TypeMap = any> {
...
+  sum(subset?: (keyof T)[], dtype?: DataType) {
+    subset     = (subset == undefined) ? this.names as (keyof T)[] : subset;
+    const temp = new Table({columns: this.select(subset)._accessor.columns});
+
+    let common_dtype: DataType = temp.getColumnByIndex(0).type;
+    const sums: number[]         = new Array(temp.numColumns);
+    for (let i = 0; i < temp.numColumns; ++i) {
+      const col    = temp.getColumnByIndex(i);
+      common_dtype = findCommonType(common_dtype, col.type);
+      sums[i]      = Number(col.sum());
```
mzegar, line 1506: In the python notebook you are able to sum `StringSeries`, however, currently, our `sum` implementation for a series is only available on `NumericSeries`. I'm assuming we want to change this? I ask because the binding is marked as complete.

mzegar:
> How does `skipna` even work? https://docs.rapids.ai/api/cudf/nightly/api.html#cudf.core.dataframe.DataFrame.sum
> 
> I attempted this example in my Jupyter notebook.
> 
> ```python
> df = cudf.DataFrame({'a': [5, None, 5]})
> print(df.sum()) # 10
> 
> df2 = cudf.DataFrame({'a': [5, None, 5]})
> print(df2.sum(False)) # 10
> ```
> 
> If one was trying to sum up a `Series` with a null in it, how would that affect the sum? Not clicking for me.

bryevdv:
@mzegar IIRC `skipna` needs to be passed as a kwarg:
```
In [3]: df.sum()
Out[3]: 
a    10
dtype: int64

In [4]: df.sum(skipna=False)
Out[4]: 
a    <NA>
dtype: float64
```
not sure offhand what a positional arg there means

mzegar:
Also need to fix up allocating the `.js` array within the method.

modules/cudf/src/data_frame.ts
```diff
@@ -1494,6 +1494,22 @@ export class DataFrame<T extends TypeMap = any> {
...
+    let common_dtype: DataType = temp.getColumnByIndex(0).type;
+    const sums: number[]       = new Array(temp.numColumns);
+    for (let i = 0; i < temp.numColumns; ++i) {
+      const col    = temp.getColumnByIndex(i);
+      common_dtype = findCommonType(common_dtype, col.type);
+      sums[i]      = Number(col.sum(memoryResource));
+    }
+
+    const result = Series.new({type: common_dtype, data: sums});
+    return dtype ? result.cast(dtype, memoryResource) : result;
```
trxcllnt, line 1510:
```suggestion
    subset = (subset == undefined) ? this.names as (keyof T)[] : subset;
    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
    const commonDtype = subset.reduce(
      (type: DataType|null, name) => findCommonType(type ?? this.types[name], this.types[name]),
      null)!;
    const sums   = subset.map((name) => this.get(name)._col.sum(memoryResource));
    const result = Series.new({type: commonDtype, data: sums});
    return dtype ? result.cast(dtype, memoryResource) : result;
```

mzegar: Neat, very similar to the stuff you did in `concat.ts`
https://github.com/rapidsai/node-rapids/blob/e5a6c44d09a95de3c3e3e42fda1446ea6b113c72/modules/cudf/src/dataframe/concat.ts#L65

modules/cudf/src/data_frame.ts
```diff
@@ -1542,6 +1542,46 @@ export class DataFrame<T extends TypeMap = any> {
...
+      null)!;
+    const sums = subset.map((name) => {
+      if (!(this.get(name) instanceof NumericSeries)) {
+        throw new Error('Unable to sum the results of non NumericSeries');
+      }
+      return Number(this.get(name)._col.sum(memoryResource));
+    });
+    let result = Series.new({type: commonDtype, data: sums});
+    if (skipna) result = result.dropNulls();
+    return dtype ? result.cast(dtype) : result;
```
mzegar, line 1582: Not sure if we want to change this up any bit, could use ternary operators to figure this out as well.

mzegar:
https://docs.rapids.ai/api/cudf/nightly/api.html#cudf.core.dataframe.DataFrame.sum

- I'm assuming `axis` is similar to `subset`, so using that instead.
- `dtype` is left out since users can just chain call `.cast()` right after
- Do we want to include `min_count`?

modules/cudf/src/data_frame.ts
```diff
@@ -1542,6 +1542,51 @@ export class DataFrame<T extends TypeMap = any> {
...
+    subset = (subset == undefined) ? this.names as (keyof T)[] : subset;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const commonDtype = subset.reduce(
+      (type: DataType|null, name) => findCommonType(type ?? this.types[name], this.types[name]),
+      null)!;
+    const sums   = subset.map((name) => {
+      if (!(this.get(name) instanceof NumericSeries)) {
+        throw new Error('Unable to sum the results of non NumericSeries');
+      }
+      return Number(this.get(name)._col.sum(memoryResource));
```
bryevdv, line 1584: Would defining `sum` on `NumericSeries` help simplify this at all?

mzegar: Chatted with Bryan, seeing if there is someway to improve the `.map` type wise.

mzegar: hmm... no luck on my end. Best I came up with is currently apart of the PR.

modules/cudf/src/data_frame.ts
```diff
@@ -1542,6 +1542,49 @@ export class DataFrame<T extends TypeMap = any> {
...
+    subset = (subset == undefined) ? this.names as (keyof T)[] : subset;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const commonDtype = subset.reduce(
+      (type: DataType|null, name) => findCommonType(type ?? this.types[name], this.types[name]),
+      null)!;
+    const sums = subset.map((name) => {
+      if (!(this.get(name) instanceof NumericSeries)) {
+        throw new Error('Unable to sum the results of non NumericSeries');
+      }
+      return Number((this.get(name) as any).sum(skipna, memoryResource));
```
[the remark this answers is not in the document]
mzegar: Casting everything to `Number` was my best guess in handling that.

```ts
Number((this.get(name) as any).sum(skipna, memoryResource))
```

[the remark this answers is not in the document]
trxcllnt: But that's a lossy conversion. We should probably throw an error or up-cast beforehand instead.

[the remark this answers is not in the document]
mzegar: Right, makes sense, could you clear up a bit how you'd see both implementations functioning?

"Throw an error"
I'm assuming you mean throw an error if one of the types inside of `sums` does not match up another type inside of `sum`?

"Up-cast beforehand"
1. We cast to a `BigInt` before adding to sums?
2. We write a method similar to `findCommonType`, but it takes in a `Number`, `BigInt`, and we properly up-cast any elements in the `sums` array.

Let me know if I'm misunderstanding.

mzegar:
Hmm, I feel like I've completely over complicated your suggestions @trxcllnt. I'm pretty unsatisfied with the current implementation in the PR. The other solutions I've thought of is...

```ts
  sum<P extends keyof T>(subset?: (keyof T)[], skipna = true, memoryResource?: MemoryResource) {
    subset = (subset == undefined) ? this.names as (keyof T)[] : subset;
    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
    let containsInt = false;
    let containsUint = false;
    ...
    const sums =
      subset.map((name) => { 
        // Do the type checks inside here on each element
        return (this.get(name) as any).sum(skipna, memoryResource); 
      });
    // do the same containsInt & containsUint, etc. checking here, throw if any errors
    return Series.new(sums) as any as Series < T[P] extends(Integral | FloatingPoint) ? T[P]
                                                                                      : never > ;
  }
  ```

This doesn't do any throwing at run-time, but instead returns the type `never` if the dataframe contains incorrect types. 
```
  sum<P extends keyof T>(subset?: (keyof T)[], skipna = true, memoryResource?: MemoryResource) {
    subset = (subset == undefined) ? this.names as (keyof T)[] : subset;
    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
    const sums =
      subset.map((name) => { return (this.get(name) as any).sum(skipna, memoryResource); });
    return Series.new(sums) as any as Series < T[P] extends(Integral | FloatingPoint) ? T[P]
                                                                                      : never > ;
  }
```

**Comment by mzegar** [comment]
I spent a decent amount of time messing with this, as it was some more time with messing with the typescript type system.

1. If we want to properly be able to throw, we will be forced to do extra iterations / checks to ensure all elements within the dataframe are either all of type Integral or Floating point. I have an example of this in the current state of this PR, it can probably be cleaned up a bit.

2. We can use the type system to return `never` on all error cases, but I think that @trxcllnt mentioned there is no way to actually get it to error at compile time.

```ts
sum<P extends keyof T>(subset?: (keyof T)[], skipna = true, memoryResource?: MemoryResource) {
  subset = (subset == undefined) ? this.names as (keyof T)[] : subset;
  const sums =
    subset.map((name) => { return (this.get(name) as any).sum(skipna, memoryResource); });
  return Series.new(sums) as any as Series < T[P] extends Integral
    ? T[P] extends FloatingPoint ? never : Integral
    : T[P] extends FloatingPoint ? FloatingPoint : never > ;
}
```

Personally, I think the second option is a lot cleaner. Thoughts?

modules/cudf/src/data_frame.ts
```diff
@@ -1588,34 +1588,31 @@ export class DataFrame<T extends TypeMap = any> {
...
+    const floatTypes = [new Float32, new Float64];
+    const intTypes   = [new Int8, new Int16, new Int32, new Int64];
+    const UintTypes  = [new Uint8, new Uint16, new Uint32, new Uint64];
+
+    const firstSeriesDtype           = this.get(subset[0]).type;
+    const isFirstSeriesFloatingPoint = floatTypes.some(t => t.compareTo(firstSeriesDtype));
+
+    const sums = subset.map((name) => {
+      if (isFirstSeriesFloatingPoint) {
+        if (!floatTypes.some(t => this.get(name).type.compareTo(t))) {
```
trxcllnt, line 1603: With the new Arrow v4.0.0 this API changed from `type.compareTo()` to a utility function `compareTypes()`: 
https://github.com/rapidsai/node-rapids/commit/1582c01381045a13f9a16a8d4c7f4db00e24d46c#diff-749cd88028943c88bd04f443e9453b52921f518db1cc04cfa0e84fa412a7a123R815

mzegar: Spoke with Paul about this, not needed since we are just going to infer using the type system.

trxcllnt (approved):

PR #221, opened 2021-06-03
[FEA] Added series `reverse`

mzegar:
Seemed reasonable to just take advantage of the spread operator, and use the built in `reverse()` in js.

### DIFF
--- a/modules/cudf/src/series.ts
+++ b/modules/cudf/src/series.ts
@@ -620,7 +620,10 @@ export class AbstractSeries<T extends DataType = any> {
    * Series.new([false, true]).reverse() // [true, false]
    * ```
    */
-  reverse(): Series<T> { return Series.new({type: this.type, data: [...this].reverse()}); }
+  reverse(): Series<T> {
+    return this.gather(
+      Series.sequence({type: new Int32, size: this.length, step: -1, init: this.length - 1}));
+  }
 
   /**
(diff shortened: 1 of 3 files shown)


modules/cudf/src/series.ts
```diff
@@ -605,6 +605,23 @@ export class AbstractSeries<T extends DataType = any> {
...
+   *
+   * // Float64Series
+   * Series.new([1, 2, 3]).reverse() // [3, 2, 1]
+   * // StringSeries
+   * Series.new(["foo", "bar"]).reverse() // ["bar", "foo"]
+   * // Bool8Series
+   * Series.new([false, true]).reverse() // [true, false]
+   * ```
+   */
+  reverse(): Series<T> { return Series.new({type: this.type, data: [...this].reverse()}); }
```
bryevdv, line 623: This will cause a copy to host, won't it? Maybe instead we could construct a reverse sequence and then gather? I think that is what Python cudf does:

https://github.com/rapidsai/cudf/blob/5b8895dc50a5daf99bafede67617e8ebd39d3af6/python/cudf/cudf/core/series.py#L3701-L3704

mzegar: spoke with Bryan a bit, cleared up allocating host memory vs. letting `CUDA/CUDF` do the work. Fixed.

bryevdv (approved):

PR #226, opened 2021-06-08
Implement cumulative reductions for Series

bryevdv:
This PR implements `cummax`, `cummin`, `cumprod`, and `cumsum` for `Series`

Currently `skipna=true` is finished, the `skipna=false` case is still TODO (some additional helpers need to be added)

### DIFF
--- a/modules/cudf/src/series/bool.ts
+++ b/modules/cudf/src/series/bool.ts
@@ -60,12 +60,12 @@ export class Bool8Series extends NumericSeries<Bool8> {
    * import {Series} from '@rapidsai/cudf';
    * const a = Series.new([true, false, true])
    *
-   * a.cummax() // {true, true, true}
+   * a.cumulativeMax() // {true, true, true}
    * ```
    */
-  cummax(skipna = true, memoryResource?: MemoryResource) {
+  cumulativeMax(skipna = true, memoryResource?: MemoryResource) {
     const result_series = this._prepare_scan_series(skipna);
-    return Series.new(result_series._col.cummax(memoryResource));
+    return Series.new(result_series._col.cumulativeMax(memoryResource));
   }
 
   /**
@@ -81,12 +81,12 @@ export class Bool8Series extends NumericSeries<Bool8> {
    * import {Series} from '@rapidsai/cudf';
    * const a = Series.new([true, false, true])
    *
-   * a.cummin() // {true, false, false}
+   * a.cumulativeMin() // {true, false, false}
    * ```
    */
-  cummin(skipna = true, memoryResource?: MemoryResource) {
+  cumulativeMin(skipna = true, memoryResource?: MemoryResource) {
     const result_series = this._prepare_scan_series(skipna);
-    return Series.new(result_series._col.cummin(memoryResource));
+    return Series.new(result_series._col.cumulativeMin(memoryResource));
   }
 
   /**
@@ -103,12 +103,12 @@ export class Bool8Series extends NumericSeries<Bool8> {
    * import {Series} from '@rapidsai/cudf';
    * const a = Series.new([true, false, true])
    *
-   * a.cumprod() // {1n, 0n, 0n}
+   * a.cumulativeProduct() // {1n, 0n, 0n}
    * ```
    */
-  cumprod(skipna = true, memoryResource?: MemoryResource) {
+  cumulativeProduct(skipna = true, memoryResource?: MemoryResource) {
     const result_series = this._prepare_scan_series(skipna).cast(new Int64, memoryResource);
-    return Series.new(result_series._col.cumprod(memoryResource));
+    return Series.new(result_series._col.cumulativeProduct(memoryResource));
   }
 
   /**
@@ -125,11 +125,11 @@ export class Bool8Series extends NumericSeries<Bool8> {
    * import {Series} from '@rapidsai/cudf';
    * const a = Series.new([true, false, true])
    *
-   * a.cumsum() // {1n, 1n, 2n}
+   * a.cumulativeSum() // {1n, 1n, 2n}
    * ```
    */
-  cumsum(skipna = true, memoryResource?: MemoryResource) {
+  cumulativeSum(skipna = true, memoryResource?: MemoryResource) {
     const result_series = this._prepare_scan_series(skipna).cast(new Int64, memoryResource);
-    return Series.new(result_series._col.cumsum(memoryResource));
+    return Series.new(result_series._col.cumulativeSum(memoryResource));
   }
 }
--- a/modules/cudf/src/series/bool.ts
+++ b/modules/cudf/src/series/bool.ts
@@ -16,7 +16,7 @@ import {Uint8ClampedBuffer} from '@nvidia/cuda';
 import {MemoryResource} from '@rapidsai/rmm';
 
 import {Series} from '../series';
-import {Bool8, Int64} from '../types/dtypes';
+import {Bool8, Int32, Int64} from '../types/dtypes';
 
 import {NumericSeries} from './numeric';
 
@@ -31,11 +31,20 @@ export class Bool8Series extends NumericSeries<Bool8> {
     return new Uint8ClampedBuffer(this._col.data).subarray(this.offset, this.offset + this.length);
   }
 
-protected _prepare_scan_series(skipna: boolean) {
-    if (skipna) { return this; }
+  protected _prepare_scan_series(skipna: boolean) {
+    if (skipna || !this.hasNulls) { return this; }
 
-    // TODO: skipna=false
-    return this;
+    const index = Series.sequence({type: new Int32, size: this.length, step: 1, init: 0});
+
+    const first = index.filter(this.isNull()).getValue(0)!;
+    const slice =
+      Series.sequence({type: new Int32, size: this.length - first, step: 1, init: first});
+
+    const copy = this.cast(this.type);
+    const mask = [...index.cast(new Bool8).fill(true).scatter(false, slice)];
+    copy.setNullMask(mask as any);
+
+    return copy;
   }
 
   /**
--- a/modules/cudf/src/series/bool.ts
+++ b/modules/cudf/src/series/bool.ts
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 import {Uint8ClampedBuffer} from '@nvidia/cuda';
+import {MemoryResource} from '@rapidsai/rmm';
 
-import {Bool8} from '../types/dtypes';
+import {Series} from '../series';
+import {Bool8, Int64} from '../types/dtypes';
 
 import {NumericSeries} from './numeric';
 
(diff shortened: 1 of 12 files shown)


modules/cudf/src/series/bool.ts
```diff
@@ -28,4 +30,97 @@ export class Bool8Series extends NumericSeries<Bool8> {
...
+
+    // TODO: skipna=false
+    return this;
+  }
+
+  /**
+   * Compute the cumulative max of all values in this Series.
+   *
+   * @param skipna The optional skipna if true drops NA and null values before computing reduction,
+   * else if skipna is false, reduction is computed directly.
```
mzegar, line 45: Our current implementations of anything related to `na` only handles `na` values.

For example, `dropNaNs` only dropping `NA` values, and `dropNulls` only handling nulls. So whenever `skipna` is handled, I think it would make sense to only remove `NA` values, and update this doc string.

modules/cudf/src/series/bool.ts
```diff
@@ -28,4 +30,97 @@ export class Bool8Series extends NumericSeries<Bool8> {
   get data() {
     return new Uint8ClampedBuffer(this._col.data).subarray(this.offset, this.offset + this.length);
   }
+
+  _prepare_scan_series(skipna: boolean) {
```
mzegar, line 34: Might be worth to set this to `true` as a default value.

mzegar (commented):
Minor initial comments

modules/cudf/src/series/bool.ts
bryevdv: I'd like to make sure an explicit argument is always passed in to help ensure that the public API value (which is optional) is always used and passed on as expected.

modules/cudf/src/series/bool.ts
bryevdv: I actually just copied this from `mean`:

https://github.com/rapidsai/node-rapids/blob/abdab8772eb0fa5f8455516f48acf714bfe22088/modules/cudf/src/series/float.ts#L167-L168

But I do think it probably could be worded better. E.g. in `float.ts` it could be "NA and NaN values", and for bool and int, it could just be "NA values". That said we do convert nulls to NA in the constructors so maybe it's good to include "null" to reflect that. Do you have thoughts or preferences @trxcllnt ?

trxcllnt: Not sure what the question is? We only have the concept of nulls (represented in the `nullMask`), and `NaN` (which are valid floating point values in the `data` buffer).

Pandas terminology is confusing here as they represent "missing" values as NaN sentinels in numeric Series (but then have no way to differentiate between true NaNs and nulls, i.e. Python's `None`), and `None` in non-numeric Series (i.e. strings).

modules/cudf/src/series/integral.ts
```diff
@@ -190,6 +190,98 @@ abstract class IntSeries<T extends Integral> extends NumericSeries<T> {
   bitInvert(memoryResource?: MemoryResource): Series<T> {
     return Series.new(this._col.bitInvert(memoryResource));
   }
+  _prepare_scan_series(skipna: boolean) {
```
trxcllnt, line 193:
```suggestion

  protected _prepare_scan_series(skipna: boolean) {
```

modules/cudf/src/series/bool.ts
bryevdv, line 34:
```suggestion
protected _prepare_scan_series(skipna: boolean) {
```

(anticipating this one)

modules/cudf/src/series/bool.ts
bryevdv: @trxcllnt would `skipNulls` be a better param name for node-rapids then? (throughout)

modules/cudf/test/series/reduce/cummax-tests.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const numbers        = [4, 2, null, 5, 1, 1];
+const float_with_NaN = [4, 2, NaN, 5, 1, 1];
+const bools          = [true, false, null, false, true];
+const bigints        = [4n, 2n, null, 5n, 1n, 1n];
+
+function testNumberAny<T extends Int8|Int16|Int32|Uint8|Uint16|Uint32|Float32|Float64>(
+  type: T, data: (T['scalarType']|null)[], skipna = true) {
+  const expected = skipna ? [4, 4, null, 5, 5, 5]  //
+                          : [4, 4, null, null, null, null];
```
bryevdv, line 44: Question here (and other tests) should the `float-nan` cases match this output exactly as well? Making that so requires calling `nansToNulls` on float series unconditionally. Python cudf on calls `nans_to_nulls` when `skipna=True` (but also seems to treat nans as nulls from the start)

modules/cudf/src/series/bool.ts
mzegar: That seems reasonable to me.

modules/cudf/test/series/reduce/cummax-tests.ts
mzegar: If I'm understanding correctly, calling `nansToNulls` unconditionally seems a bit odd to me. Be curious to hear what others think.

bryevdv: well, in a nutshell, this difference is why:

### Python cudf
```
In [3]: a = cudf.Series([4, 2, np.NaN, 5, 1, 1.0])

In [4]: a
Out[4]: 
0     4.0
1     2.0
2    <NA>
3     5.0
4     1.0
5     1.0
dtype: float64

In [5]: a.has_nulls
Out[5]: True
```

#### node-rapids cudf
```
> a = Series.new([2, 1, NaN, 3, 4.0])
Float64Series { _col: Column {} }
> a.hasNulls
false
```

---

To be honest, given the above I am not sure why Python calls `nans_to_nulls` at all. Maybe it is just cruft let over from the past?

modules/cudf/src/series/bool.ts
```diff
@@ -28,4 +30,106 @@ export class Bool8Series extends NumericSeries<Bool8> {
...
+   * else if skipna is false, reduction is computed directly.
+   * @param memoryResource The optional MemoryResource used to allocate the result Series's device
+   *   memory.
+   * @returns The cumulative product of all the values in this Series.
+   * @example
+   * ```typescript
+   * import {Series} from '@rapidsai/cudf';
+   * const a = Series.new([true, false, true])
+   *
+   * a.cumprod() // {1n, 1n, 2n}
```
mzegar, line 106: Should these results be `[1n, 0n, 0n]`?

Also, extremely nit picky, I think most of our docs use `[` instead of `{`

modules/cudf/src/series/bool.ts
```diff
@@ -28,4 +30,106 @@ export class Bool8Series extends NumericSeries<Bool8> {
...
+   * else if skipna is false, reduction is computed directly.
+   * @param memoryResource The optional MemoryResource used to allocate the result Series's device
+   *   memory.
+   * @returns The cumulative sum of all the values in this Series.
+   * @example
+   * ```typescript
+   * import {Series} from '@rapidsai/cudf';
+   * const a = Series.new([true, false, true])
+   *
+   * a.cumsum() // {1n, 0n, 0n}
```
mzegar, line 128: And this should be `[1n, 1n, 2n]`

modules/cudf/src/series/integral.ts
```diff
@@ -190,6 +191,108 @@ abstract class IntSeries<T extends Integral> extends NumericSeries<T> {
...
+   * else if skipna is false, reduction is computed directly.
+   * @param memoryResource The optional MemoryResource used to allocate the result Series's device
+   *   memory.
+   * @returns The cumulative max of all the values in this Series.
+   * @example
+   * ```typescript
+   * import {Series} from '@rapidsai/cudf';
+   * const a = Series.new([4, 2, 5, 1, 1])
+   *
+   * a.cummax() // {4, 4, 5, 5, 5}
```
mzegar, line 224: Should these values inside the array be big ints? i.e `4` -> `4n`?

mzegar (commented):
Got some questions/comments.

modules/cudf/src/series/bool.ts
bryevdv: I picked the `{` up from the libcudf docs, I think it's a nice way to indicate the answer without suggesting that the value is a literal JS array, but I don't have strong feelings here

bryevdv, line 106:
```suggestion
   * a.cumprod() // {1n, 0n, 0n}
```

modules/cudf/src/series/bool.ts
bryevdv, line 128:
```suggestion
   * a.cumsum() // {1n, 1n, 2n}
```

modules/cudf/src/series/bool.ts
mzegar: That seems reasonable, no hard preference either.

modules/cudf/src/series/integral.ts
bryevdv: bigint output only for bigint input AFAIK. That said, `Series.new([4, 2, 5, 1, 1])` actually creates a Float32 series. It would be more precise but tedious to to`Series.new({type: new Int32, data: [4, 2, 5, 1, 1]})` which would yield the current output. 

Or, change to `a = Series.new([4n, 2n, 5n, 1n, 1n])` input to get the bigint output you suggest. 

@trxcllnt do you have a preference on which to demo in the docstrings? (or both?)

modules/cudf/src/series/bool.ts
bryevdv: @AjayThorve @trxcllnt do either of yall have a strong preference?

modules/cudf/src/series/integral.ts
mzegar: Ah right, that makes sense. Thanks.

modules/cudf/src/series/bool.ts
trxcllnt: @bryevdv yeah I definitely think that would reduce confusion

bryevdv: :+1: since it will affect files otherwise outside the scope of this PR, I'll open a new issue.

modules/cudf/src/series/integral.ts
bryevdv: per @trxcllnt just adding `.cast(new Int32)` is also nice

modules/cudf/test/series/reduce/cummax-tests.ts
bryevdv: Maybe this question is better to raise in a dedicated issue?

bryevdv:
@AjayThorve @trxcllnt @mzegar Any thoughts on renaming these to have more verbose names? 

* `cumulativeSum`
* `cumulativeProduct`
* `cumulativeMax`
* `cumulativeMin`

I suppose `runningSum` etc could also work.

mzegar:
> @AjayThorve @trxcllnt @mzegar Any thoughts on renaming these to more verbose names?
> 
> * `cumulativeSum`
> * `cumulativeProduct`
> * `cumulativeMax`
> * `cumulativeMin`
> 
> I suppose `runningSum` etc could also work. Given how nicely vscode and other IDEs auto-complete, I am not sure why we should need to keep the abbreviated names.

I personally like the verbose names.

AjayThorve:
> @AjayThorve @trxcllnt @mzegar Any thoughts on renaming these to more verbose names?
> 
> * `cumulativeSum`
> * `cumulativeProduct`
> * `cumulativeMax`
> * `cumulativeMin`
> 
> I suppose `runningSum` etc could also work. Given how nicely vscode and other IDEs auto-complete, I am not sure why we should need to keep the abbreviated names.

Only argument against verbose names would be keeping it inline with the python api? But then again, we have diverged from python api's in other cases, so I'm fine with the verbose names.

bryevdv:
OK I have updated the example code and also changed to the more verbose names. cc @AjayThorve @mzegar

mzegar (approved):
LGTM

trxcllnt:
@bryevdv I like the longer names!

PR #228, opened 2021-06-10
[BUG] Fix `Column.setNullMask`, add tests, move Column buffer helpers to utilities

trxcllnt:
PR to make `setNullMask()` work as expected.

### DIFF
--- a/modules/cudf/test/cudf-column-tests.ts
+++ b/modules/cudf/test/cudf-column-tests.ts
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Float32Buffer, Int32Buffer, setDefaultAllocator, Uint8Buffer} from '@nvidia/cuda';
+import {
+  DeviceMemory,
+  Float32Buffer,
+  Int32Buffer,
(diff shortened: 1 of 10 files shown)


modules/cudf/test/cudf-column-tests.ts
```diff
@@ -143,3 +149,63 @@ test('Column.nansToNulls', () => {
...
+    validateSetNullMask(makeTestColumn(4), 4, 4, [0, 0, 0, 0]);
+  });
+
+  test('accepts Arrays of bigints', () => {
+    validateSetNullMask(makeTestColumn(4), 4, 4, [0n, 0n, 0n, 0n]);
+  });
+
+  test('accepts Arrays of booleans', () => {
+    validateSetNullMask(makeTestColumn(4), 4, 4, [false, false, false, false]);
+  });
```
bryevdv, line 203: Maybe good to explicitly test a not-all-null case and the no-nulls case with array inputs?

bryevdv (approved):

bryevdv (approved):

AjayThorve (approved):

PR #229, opened 2021-06-14
Add Series.skew and Series.kurtosis

bryevdv:
Quick PR to add `Series.skew` and `Series.kurtosis`.

### DIFF
--- a/modules/cudf/src/series/numeric.ts
+++ b/modules/cudf/src/series/numeric.ts
@@ -47,6 +47,9 @@ export abstract class NumericSeries<T extends Numeric> extends Series<T> {
     return Series.new(this._col.cast(dataType, memoryResource));
   }
 
+  /** @ignore */
+  nansToNulls(_memoryResource?: MemoryResource): Series<T> { return this.__construct(this._col); }
+
   /**
    * View the data underlying this Series as a new dtype (similar to `reinterpret_cast` in C++).
    *
@@ -1202,10 +1205,6 @@ export abstract class NumericSeries<T extends Numeric> extends Series<T> {
     return Series.new(this._col.not(memoryResource));
   }
 
-  _process_reduction(skipna = true, memoryResource?: MemoryResource): Series<T> {
-    return skipna ? this.dropNulls(memoryResource) : this.__construct(this._col);
-  }
-
   /**
    * Compute the min of all values in this Column.
    * @param skipna The optional skipna if true drops NA and null values before computing reduction,
@@ -1221,7 +1220,8 @@ export abstract class NumericSeries<T extends Numeric> extends Series<T> {
    * a.min() // [1]
    */
   min(skipna = true, memoryResource?: MemoryResource) {
-    return this._process_reduction(skipna, memoryResource)._col.min(memoryResource);
+    const data = skipna ? this.nansToNulls(memoryResource).dropNulls(memoryResource) : this;
+    return data._col.min(memoryResource);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


modules/cudf/src/series/numeric.ts
```diff
@@ -1407,6 +1407,82 @@ export abstract class NumericSeries<T extends Numeric> extends Series<T> {
...
+   * import {Series} from '@rapidsai/cudf';
+   * const a = Series.new([1, 2, 3, 4, 5, 6, 6]);
+   *
+   * a.skew() // -0.288195490292614
+   * ```
+   */
+  skew(skipna = true, memoryResource?: MemoryResource) {
+    if (this.length == 0 || (this.hasNulls && !skipna)) { return NaN; }
+
+    const data = this._process_reduction(skipna, memoryResource);
```
bryevdv, line 1470: This seems like a poorly named function. It's basically `this.nansToNulls().dropNulls()` or float series and `this.dropNulls()` for everything else. Should we consider letting `nansToNulls` exist on integral series, but be a no-op? 

This is certainly what Python cudf does:
```
In [45]: a = cudf.Series([1,2, 3])

In [47]: a.nans_to_nulls()
Out[47]: 
0    1
1    2
2    3
dtype: int64
```

modules/cudf/src/series/numeric.ts
```diff
@@ -1407,6 +1407,82 @@ export abstract class NumericSeries<T extends Numeric> extends Series<T> {
...
+    const n = data.length;
+    if (n < 4) { return NaN; }
+
+    const V = data.var(skipna, 1, memoryResource);  // ddof = 1
+    if (V == 0) { return 0; }
+
+    const mu = data.mean(skipna, memoryResource);
+
+    const m4 =
+      Number(data.sub(mu, memoryResource).pow(4, memoryResource).sum(skipna, memoryResource)) /
```
mzegar, line 1441: After similarly using the `Number(...)` constructor in my https://github.com/rapidsai/node-rapids/pull/220 PR, Paul pointed out that there would be information loss if we were to convert a `bigint` to a `number`. 

Is this a concern in `kurtosis` and `skew`? I know `mean()` returns a number, so unless you upcast (if you even can), you wouldn't be able to do the division on line 1441.

mzegar (commented):
Just a quick question/comment.

modules/cudf/src/series/numeric.ts
AjayThorve: yep, I think that might be more consistent. +1 for a no-op `nansToNulls` operation for integral.

mzegar: Agreed, seems reasonable.

modules/cudf/src/series/numeric.ts
bryevdv: I'm not sure about upcasting. When the values are actually big, then the results seem reasonable (tho different) either way:
```
> BigInt(1000000000000000000000000000000000000000000000000) / BigInt(3)
333333333333333347948614348358732451544682550613n

> Number(BigInt(1000000000000000000000000000000000000000000000000)) / 3
3.3333333333333336e+47
```

However if you are dealing with "BigInt" values that are actually not, in fact, very big for whatever reason, then the upcasting results do not seem OK at all:
```
> Number(BigInt(10)) / 3
3.3333333333333335
> BigInt(10) /  BigInt(3)
3n
```

Definitely would like to hear from @trxcllnt in case there's just some other option I'm not aware of

AjayThorve (approved):
LGTM

modules/cudf/src/series/numeric.ts
bryevdv: left that for a future issue/PR https://github.com/rapidsai/node-rapids/issues/231

modules/cudf/src/series/numeric.ts
bryevdv: handled per discussion w/ @trxcllnt

bryevdv:
@AjayThorve those changes are done

PR #236, opened 2021-06-18
[FEA] Add `nlargest`, `nsmallest` series binding

mzegar:
https://github.com/rapidsai/cudf/blob/90fe10b2947c9caee998f6236d432f3742c3504e/python/cudf/cudf/core/series.py#L3242

### DIFF
--- a/modules/cudf/src/series.ts
+++ b/modules/cudf/src/series.ts
@@ -681,8 +681,15 @@ export class AbstractSeries<T extends DataType = any> {
    * a.nLargest(-1); // []
    * ```
    */
-  nLargest(n = 5, keep: DuplicateKeepOption = DuplicateKeepOption.first) {
-    return this._nLargestOrSmallest(true, n, keep);
+  nLargest(n = 5, keep: keyof typeof DuplicateKeepOption = 'first') {
+    if (keep == 'first') {
+      return this.sortValues(false).head(n < 0 ? 0 : n);
+    } else if (keep == 'last') {
+      return n <= 0 ? Series.new({type: this.type, data: new Array(0)})
+                    : this.sortValues(true).tail(n).reverse();
+    } else {
+      throw new TypeError('keep must be either "first" or "last"');
+    }
   }
 
   /**
@@ -703,8 +710,15 @@ export class AbstractSeries<T extends DataType = any> {
    * a.nSmallest(-1); // []
    * ```
    */
-  nSmallest(n = 5, keep: DuplicateKeepOption = DuplicateKeepOption.first) {
-    return this._nLargestOrSmallest(false, n, keep);
+  nSmallest(n = 5, keep: keyof typeof DuplicateKeepOption = 'first') {
+    if (keep == 'first') {
+      return this.sortValues(true).head(n < 0 ? 0 : n);
+    } else if (keep == 'last') {
+      return n <= 0 ? Series.new({type: this.type, data: new Array(0)})
+                    : this.sortValues(false).tail(n).reverse();
+    } else {
+      throw new TypeError('keep must be either "first" or "last"');
+    }
   }
 
   /**
--- a/modules/cudf/src/series.ts
+++ b/modules/cudf/src/series.ts
@@ -673,7 +673,7 @@ export class AbstractSeries<T extends DataType = any> {
       return n <= 0 ? Series.new({type: this.type, data: new Array(0)})
                     : this.sortValues(true).tail(n).reverse();
     } else {
-      throw new Error('keep must be either "first" or "last"');
+      throw new TypeError('keep must be either "first" or "last"');
     }
   }
 
--- a/modules/cudf/src/series.ts
+++ b/modules/cudf/src/series.ts
@@ -649,9 +649,24 @@ export class AbstractSeries<T extends DataType = any> {
   }
 
   /**
+   * perform nLargest or nSmallest operation
+   * @ignore
+   */
+  _nLargestOrSmallest(ascending: boolean, n: number, keep: DuplicateKeepOption) {
+    if (keep == DuplicateKeepOption.first) {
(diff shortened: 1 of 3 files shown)


modules/cudf/src/series.ts
```diff
@@ -648,6 +648,65 @@ export class AbstractSeries<T extends DataType = any> {
     return this.__construct(this._col.gather(selection._col));
   }
 
+  /**
+   * perform nLargest or nSmallest operation
+   * @ignore
+   */
+  _nLargestOrSmallest(ascending: boolean, n: number, keep: DuplicateKeepOption) {
```
trxcllnt, line 655:
```suggestion
  protected _nLargestOrSmallest(ascending: boolean, n: number, keep: DuplicateKeepOption) {
```

modules/cudf/src/series.ts
```diff
@@ -648,6 +648,65 @@ export class AbstractSeries<T extends DataType = any> {
...
+    if (keep == DuplicateKeepOption.first) {
+      return this.sortValues(!ascending).head(n < 0 ? 0 : n);
+    } else if (keep == DuplicateKeepOption.last) {
+      return n <= 0 ? Series.new({type: this.type, data: new Array(0)})
+                    : this.sortValues(ascending).tail(n).reverse();
+    } else {
+      throw new TypeError('keep must be either "first" or "last"');
+    }
+  }
```
trxcllnt, line 664: Why not have these logic branches in each of the functions directly?

trxcllnt (changes requested):
Left a question

modules/cudf/src/series.ts
```diff
@@ -648,6 +648,65 @@ export class AbstractSeries<T extends DataType = any> {
...
+   * const b = Series.new(["foo", "bar", "test"]);
+   *
+   * a.nLargest(); // [12, 10, 8, 6, 4]
+   * b.nLargest(1); // ["test"]
+   * a.nLargest(-1); // []
+   * ```
+   */
+  nLargest(n = 5, keep: DuplicateKeepOption = DuplicateKeepOption.first) {
+    return this._nLargestOrSmallest(true, n, keep);
+  }
```
trxcllnt, line 686:
```suggestion
  nLargest(n = 5, keep: keyof typeof DuplicateKeepOption = 'first') {
    return this._nLargestOrSmallest(true, n, DuplicateKeepOption[keep]);
  }
```

modules/cudf/src/series.ts
```diff
@@ -648,6 +648,65 @@ export class AbstractSeries<T extends DataType = any> {
...
+   * const b = Series.new(["foo", "bar", "test"]);
+   *
+   * a.nSmallest(); // [1, 2, 4, 6, 8]
+   * b.nSmallest(1); // ["bar"]
+   * a.nSmallest(-1); // []
+   * ```
+   */
+  nSmallest(n = 5, keep: DuplicateKeepOption = DuplicateKeepOption.first) {
+    return this._nLargestOrSmallest(false, n, keep);
+  }
```
trxcllnt, line 708:
```suggestion
  nSmallest(n = 5, keep: keyof typeof DuplicateKeepOption = 'first') {
    return this._nLargestOrSmallest(false, n, DuplicateKeepOption[keep]);
  }
```

modules/cudf/src/series.ts
mzegar: Do you mean...

```ts
nLargest(n = 5, keep: DuplicateKeepOption = DuplicateKeepOption.first) {
    if (keep == DuplicateKeepOption.first) {
      return this.sortValues(!ascending).head(n < 0 ? 0 : n);
    } else if (keep == DuplicateKeepOption.last) {
      return n <= 0 ? Series.new({type: this.type, data: new Array(0)})
                    : this.sortValues(ascending).tail(n).reverse();
    } else {
      throw new TypeError('keep must be either "first" or "last"');
    }
}

nSmallest(n = 5, keep: DuplicateKeepOption = DuplicateKeepOption.first) {
    if (keep == DuplicateKeepOption.first) {
      return this.sortValues(!ascending).head(n < 0 ? 0 : n);
    } else if (keep == DuplicateKeepOption.last) {
      return n <= 0 ? Series.new({type: this.type, data: new Array(0)})
                    : this.sortValues(ascending).tail(n).reverse();
    } else {
      throw new TypeError('keep must be either "first" or "last"');
    }
}

```
(ignoring any syntax errors / incorrect variables).

I basically set it up as it is in `CUDF` https://github.com/rapidsai/cudf/blob/90fe10b2947c9caee998f6236d432f3742c3504e/python/cudf/cudf/core/series.py#L3242

I don't really have a preference, I'm fine to change it.

modules/cudf/src/series.ts
mzegar: What's the reasoning behind preferring this over the enum? Keeping it similar to how it's used in python?

trxcllnt: Yeah that, and to match how we're doing it in other places

modules/cudf/src/series.ts
trxcllnt: Oh my bad, I didn't realize the switching on `keep` wasn't about the largest/smallest, but the first/last.

modules/cudf/src/series.ts
trxcllnt: We could also just make this a utility function at the bottom of the file. Then it's not on the `Series` prototype when people look at series instances in the debugger.

trxcllnt (commented):
I was wrong, I think you had it right the first time :sweat_smile:

mzegar:
> I was wrong, I think you had it right the first time :sweat_smile: 

Np, I'll go with the utility function at the bottom of the file.

trxcllnt (approved):

PR #239, opened 2021-06-22
[FEA] `Kurtosis` and `skew` for dataframe

mzegar:
From my testing, this should always return `never` if provided an invalid DataFrame containing non numeric Series.

### DIFF
--- a/modules/cudf/test/cudf-data-frame-tests.ts
+++ b/modules/cudf/test/cudf-data-frame-tests.ts
@@ -725,6 +725,30 @@ describe('dataframe.replaceNulls', () => {
   });
 });
 
+test('dataframe.kurtosis', () => {
+  const a  = Series.new([1, 2, 3, 4]);
+  const b  = Series.new([7, 8, 9, 10]);
+  const df = new DataFrame({'a': a, 'b': b});
+
+  expect([...df.kurtosis()]).toEqual([-1.1999999999999904, -1.2000000000000686]);
+
+  const c        = Series.new(['foo', 'bar', 'foo', 'bar']);
+  const never_df = new DataFrame({'a': a, 'b': b, 'c': c});
+  _verifySeriesIsNever(never_df.kurtosis());
+});
+
+test('dataframe.skew', () => {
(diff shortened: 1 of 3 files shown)


modules/cudf/test/cudf-data-frame-tests.ts
```diff
@@ -832,3 +856,8 @@ test(`DataFrame.dropDuplicates("first", true, true, ['a'])`, () => {
   expect([...result.get('a')]).toEqual([null, 1, 3, 4]);
   expect([...result.get('b')]).toEqual([null, 5, 8, 2]);
 });
+
+// Typescript does not allow us to throw a compile-time error if
+// the result of a method is `never`. Instead, let's just verify
+// the result is `never`.
+function _verifySeriesIsNever(_: never) { return _; }
```
AjayThorve, line 863: shouldn't this be `_verifySeriesIsNever(_: Series<never>)` ?

mzegar: I need to double check the difference, but technically what I return is just `never`, whereas this https://github.com/rapidsai/node-rapids/pull/240 has `Dataframe<never>`. Should check out the difference.

AjayThorve: I was referring to the result of `skew` [here](https://github.com/mzegar/node-rapids/blob/df-skew-kurtosis/modules/cudf/src/data_frame.ts#L1628), where you were returing `Series<never>`.

But yes, if #240 works, then we can ignore this

mzegar: ah you're right, I'll double check this tomorrow. For whatever reason `ts` was inferring it as `never`, I'll look into this.

AjayThorve (approved):
LGTM

modules/cudf/test/cudf-data-frame-tests.ts
mzegar: Verified that this returns `never`.

PR #240, opened 2021-06-22
[FEA] Type inferred unaryops

mzegar:
`CUDF` actually throws if the `df.cos()` receives a Series that isn't of type numeric.
```python
import cudf
df = cudf.DataFrame({'a': [3, 2, 3, 4], 'b': [4, 3, 2, 4], 'c': ["foo", "bar", "foo", "bar"]})
print(df.cos()) # error
```

This change now makes it so that if any of the `Series` inside of the `DataFrame` are not of type `Numeric`, the return type is `never`. Previously, we would just skip over and ignore any `Series` that isn't type `Numeric`. So this change brings it closer in line with `CUDF`.

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -91,6 +91,19 @@ const NumericTypes = [
   new Uint64,
 ];
 
+const IntegralTypes = [
+  new Int8,
+  new Int16,
+  new Int32,
+  new Int64,
+  new Uint8,
+  new Uint16,
+  new Uint32,
+  new Uint64,
+];
+
+const FloatTypes = [new Float32, new Float64];
+
 function _seriesToColumns<T extends TypeMap>(data: SeriesMap<T>) {
   const columns = {} as any;
   for (const [name, series] of Object.entries(data)) { columns[name] = series._col; }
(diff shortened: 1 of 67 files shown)


AjayThorve:
definitely prefer this way over the current version. Also, not sure if we skipped over non-numeric series in the current version, as it seems to include the StringSeries in the return hints as well

image

mzegar:
> definitely prefer this way over the current version. Also, not sure if we skipped over non-numeric series in the current version, as it seems to include the StringSeries in the return hints as well
> 
> image

Yeah we skipped over any non numeric series and kept it in the dataframe (in the current implemention)

modules/cudf/src/data_frame.ts
```diff
@@ -1022,16 +1015,10 @@ export class DataFrame<T extends TypeMap = any> {
...
-        ...map,
-        [name]: _invokeIfNumericSeries(
-          this.get(name),
-          () => { return Series.new(this._accessor.get(name).sin(memoryResource)); })
-      };
-    };
-    return new DataFrame(this.names.reduce(mapper, {} as SeriesMap<T>));
+  sin<P extends keyof T>(memoryResource?: MemoryResource) {
+    return new DataFrame(this.names.reduce(
+             (map, name) => ({...map, [name]: (this.get(name) as any).sin(memoryResource)}),
```
trxcllnt, line 1020: Should we throw a more descriptive error here if the types are wrong? Right now it'll throw something like `TypeError: this.get(...).sin is not a function`

trxcllnt (changes requested):
Probably good to add more descriptive error messages at runtime.

modules/cudf/src/data_frame.ts
mzegar: Spoke with Paul, I assumed that `typescript` would be a requirement to use `node-rapids`, but this isn't the case.

Regardless, this PR should provide better typing if people do decide to use typescript (i.e return `never` if the result won't occur), but also giving useful `throw` messages as well.

modules/cudf/src/data_frame.ts
```diff
@@ -71,16 +77,43 @@ type JoinProps<
...
+  new Int16,
+  new Int32,
+  new Int64,
+  new Uint8,
+  new Uint16,
+  new Uint32,
+  new Uint64,
+];
+
+const FloatTypes = [new Float32, new Float64];
```
mzegar, line 105: Do we maybe want to move these to `dtypes.ts` and export them?

trxcllnt: Yeah I think that'd be good.

trxcllnt (approved):
LGTM. I don't think we should block this on moving the dtypes, but feel free if you want to.

mzegar:
I'll create an issue for this.

PR #243, opened 2021-06-24
[FEA] Swap out `ccache` for `sccache`

trxcllnt:
Swaps out `ccache` for `sccache`, allowing us to share the common build cache via something like AWS S3 or GCP.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -12,5 +12,14 @@ CUDA_VERSION=11.2.2
 # `nvidia/cudagl` base image Ubuntu version
 LINUX_VERSION=ubuntu20.04
 
-# Path where cuDF should store its runtime-compiled JIT kernels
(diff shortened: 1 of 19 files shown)


ajschmidt8 (approved):
nice! will this require us to manage/distribute AWS credentials to node-rapids users?

mzegar (approved):
Really cool stuff, didn't find anything on my end.

AjayThorve (approved):
LGTM!!! yay for shareable build caches!!

bryevdv (approved):

love-lena (approved):
No problems in the parts I have context on.
I would like to hear more about the AWS implementation but it looks like sccache is set up to default to a local cache if no creds are provided, so shouldn't be a problem

PR #244, opened 2021-06-24
Convert strict not equals to not equals in tests

love-lena:
A strict check for null was causing an undefined value to be
passed into a function that could not handle an undefined value.

------
This change is trivial and does not completely fix the tests that Paul and I were debugging. But I wanted to get a PR submitted and approved so I have the process down.


trxcllnt (approved):
LGTM!

PR #248, opened 2021-06-25
[FIX] Change df `concat` result type to `never`

mzegar:
This is how it previously used to return `never`

```ts
const result = df1.concat(df2);
// const result: DataFrame<{
//     a: never;
// }>

// IDE now autocomplete and allows calling of base level df methods
result.isNaN(); // no compiler error
```

```ts
const result = df1.concat(df2);
// const result: never

result.isNaN(); // compiler error since this entire obj is null
```

### DIFF
--- a/modules/cudf/src/dataframe/concat.ts
+++ b/modules/cudf/src/dataframe/concat.ts
@@ -107,7 +107,13 @@ export function concat<TFirst extends DataFrame, TRest extends DataFrame[]>(firs
 
   type TResultTypeMap = ConcatTypeMap<TFirst, TRest>;
 
-  return new DataFrame(names.reduce(
-    (map, name, index) => ({...map, [name]: Series.new(concatenatedTable.getColumnByIndex(index))}),
-    {} as SeriesMap<{[P in keyof TResultTypeMap]: TResultTypeMap[P]}>));
+  // clang-format off
+  return new DataFrame(
+    names.reduce((map, name, index) =>
+    ({...map, [name]: Series.new(concatenatedTable.getColumnByIndex(index))}),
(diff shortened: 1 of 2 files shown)


modules/cudf/src/dataframe/concat.ts
```diff
@@ -107,7 +107,13 @@ export function concat<TFirst extends DataFrame, TRest extends DataFrame[]>(firs
...
-    {} as SeriesMap<{[P in keyof TResultTypeMap]: TResultTypeMap[P]}>));
+  // clang-format off
+  return new DataFrame(
+    names.reduce((map, name, index) =>
+    ({...map, [name]: Series.new(concatenatedTable.getColumnByIndex(index))}),
+    {} as SeriesMap<{[P in keyof TResultTypeMap]: TResultTypeMap[P]}>)
+  ) as TResultTypeMap[keyof TResultTypeMap] extends never
+    ? never
+    : DataFrame<{[P in keyof TResultTypeMap]: TResultTypeMap[P]}>;
+  // clang-format on
```
mzegar, line 118: Do we want to store `{[P in keyof TResultTypeMap]: TResultTypeMap[P]}` as a type? then use it on both line 114 and line 117? That's about it in terms of trying to condense this down, not sure if there are any other neat treats we can use here.

trxcllnt: The type `{[P in keyof TResultTypeMap]: TResultTypeMap[P]}` seems fine to me here as-is.

We throw an error in the case the column types can't be concatenated, right?

mzegar: Discussed this, findcommontype already throws.

trxcllnt (approved):

Issue #231, opened 2021-06-15
Rename skipna to skipNulls


bryevdv:
See discussion here: https://github.com/rapidsai/node-rapids/pull/226#discussion_r648686298

PR #249, opened 2021-06-25
Rename skipna to skipNulls

love-lena:
Renamed for clarity, no functional change

Resolves: #231

-----
Builds and passes MOST tests. There is an issue with the rmm tests on a few machines, but the change introduces no additional failures, so I consider it well enough tested. If anyone with working tests wants to verify it passes all tests be my guest, but there should be zero functionality change, this was just a ctrl+r change.


trxcllnt (approved):
LGTM

mzegar (approved):
LGTM

PR #250, opened 2021-06-26
Add DataFrame.fromArrow

bryevdv:
Add a static `DataFrame.fromArrow` method that can adapt Arrow tables stored in-memory in Arrow IPC format.
 
This is still WIP but early comments welcome. Some notes:

* As implemented, this requires a copy. Plan to to push improvements upstream to `cudf::from_arrow` to allow it to perform zero-copy when feasible (e.g. ipc buffers)
* Tests are a bit thin. Open to suggestions to improve coverage parsimoniously 
* ~~Still need to add an input type to `fromArrow` (similar to `readCSV`) so that full types can be deduced~~

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {MemoryResource} from '@rapidsai/rmm';
+import {Memory} from '@nvidia/cuda';
+import {DeviceBuffer, MemoryResource} from '@rapidsai/rmm';
 import * as arrow from 'apache-arrow';
 import {compareTypes} from 'apache-arrow/visitor/typecomparator';
 import {Readable} from 'stream';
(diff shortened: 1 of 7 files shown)


modules/cudf/src/data_frame.ts
```diff
@@ -103,6 +104,18 @@ export class DataFrame<T extends TypeMap = any> {
                    {} as ColumnsMap<{[P in keyof T]: CSVToCUDFType<T[P]>}>)));
   }
 
+  /**
+   * Reads a an arrow IPC table into a DataFrame.
```
mzegar, line 108: Grammar here `Reads a an arrow` -> `Reads an arrow`

modules/cudf/src/table.ts
```diff
@@ -45,6 +47,14 @@ interface TableConstructor {
   readCSV<T extends CSVTypeMap = any>(options: ReadCSVOptions<T>):
     {names: (keyof T)[], table: Table};
 
+  /**
+   * Reads a an arrow IPC table into a set of columns.
```
mzegar, line 51: Same as comment above.

modules/cudf/src/table/arrow.cpp
```diff
@@ -57,4 +58,41 @@ Napi::Value Table::to_arrow(Napi::CallbackInfo const& info) {
...
+  std::shared_ptr<arrow::Table> arrow_table{};
+  auto status = stream_reader->ReadAll(&arrow_table);
+
+  if (!status.ok()) { NAPI_THROW(Napi::Error::New(env, status.message())); }
+
+  auto output = Napi::Object::New(env);
+
+  auto fields = stream_reader->schema()->fields();
+  auto names  = Napi::Array::New(env, fields.size());
+  for (std::size_t i = 0; i < fields.size(); ++i) { names.Set(i, fields[i]->name()); }
```
mzegar, line 87: Is there a `std` operations for this? `transform`? Regardless, this still looks clean.

mzegar (approved):
Initial comments

modules/cudf/src/table/arrow.cpp
bryevdv: We would need an NAPI approach that didn't require the iteration index, or `std` operation that provides one (i.e. like Python's `enumerate`) but AFAIK I don't think either exists.

mzegar:
Reviewed in the morning and didn't realize I accidently approved. Will give it another look over.

modules/cudf/src/table/arrow.cpp
trxcllnt: It'd be nice if `Napi::Array` had an iterator similar to https://github.com/nodejs/node-addon-api/pull/930

mzegar:
`BINDINGS.md` needs to be updated, otherwise, got nothing else to comment Might be good to get another review :+1:

bryevdv:
Good catch, updated!

PR #257, opened 2021-06-30
Fix memory allocation bug in cudf column tests

love-lena:
---------
Fixes a failing test on only lower memory devices by limiting the amount of memory tests allocate

change written by @bryevdv


bryevdv (approved):

bryevdv:
> Fixes a failing test on only lower memory devices by limiting the amount of memory tests allocate

Just FYI: these tests were failing on the workstations too, the main issue was that the mask buffers were not zero-initialized, which caused unpredictable results

PR #264, opened 2021-07-03
[FEA] Remove esm, add npm bin files to all demos

trxcllnt:
* Removes `esm` dependency because `@babel/register` seems to do it all now
* Adds a `"bin"` entry to each demo so they can be launched like any other npm bin script

Now demos can be launched directly with `yarn demo-graph` or `npx demo-graph`.

### DIFF
--- a/modules/demo/spatial/data.js
+++ b/modules/demo/spatial/data.js
@@ -27,7 +27,7 @@ if (require.main === module) {
 
 async function loadSpatialDataset() {
 
-  const points = (await loadFiles()).reduce((points, file) => {
+  const points = (await loadFiles(1)).reduce((points, file) => {
     return points ? points.concat(reshape(file)) : reshape(file);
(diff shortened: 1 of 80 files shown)


modules/demo/spatial/data.js
```diff
@@ -83,6 +83,9 @@ function loadFile(i) {
 
       let body = new Uint8Array(res.headers['content-length'] | 0);
 
+      console.log('content-encoding:', encoding);
+      console.log('content-length:', body.byteLength);
```
mzegar, line 87: I'm assuming we are keeping this logging in on purpose?

mzegar (approved):
One potential thing.

modules/demo/spatial/data.js
trxcllnt: oh woops, nope we don't need that

Issue #260, opened 2021-07-02
Series.new should handle typed array and buffer inputs


bryevdv:
##### Typed Array input
```
> {Series} = require("@rapidsai/cudf")

> Series.new(new Int32Array([1, 2, 3]))
Uncaught Error: Unrecognized type 'NONE'
    at inferDType (/opt/node-rapids/node_modules/apache-arrow/visitor.js:276:11)
    at getVisitFn (/opt/node-rapids/node_modules/apache-arrow/visitor.js:63:17)
    at ArrowToCUDFTypeVisitor.getVisitFn (/opt/node-rapids/node_modules/apache-arrow/visitor.js:32:16)
    at ArrowToCUDFTypeVisitor.getVisitFn (/opt/node-rapids/modules/cudf/build/js/types/mappings.js:44:30)
    at ArrowToCUDFTypeVisitor.visit (/opt/node-rapids/node_modules/apache-arrow/visitor.js:29:21)
    at Object.arrowToCUDFType (/opt/node-rapids/modules/cudf/build/js/types/mappings.js:97:24)
    at asColumn (/opt/node-rapids/modules/cudf/build/js/series.js:803:33)
    at Function.new (/opt/node-rapids/modules/cudf/build/js/series.js:33:31)
```

##### Buffer input
```
> {Series} = require("@rapidsai/cudf")
> {Int32Buffer} = require("@nvidia/cuda")

> Series.new(new Int32Buffer([1, 2, 3]))
Uncaught Error: Unrecognized type 'NONE'
    at inferDType (/opt/node-rapids/node_modules/apache-arrow/visitor.js:276:11)
    at getVisitFn (/opt/node-rapids/node_modules/apache-arrow/visitor.js:63:17)
    at ArrowToCUDFTypeVisitor.getVisitFn (/opt/node-rapids/node_modules/apache-arrow/visitor.js:32:16)
    at ArrowToCUDFTypeVisitor.getVisitFn (/opt/node-rapids/modules/cudf/build/js/types/mappings.js:44:30)
    at ArrowToCUDFTypeVisitor.visit (/opt/node-rapids/node_modules/apache-arrow/visitor.js:29:21)
    at Object.arrowToCUDFType (/opt/node-rapids/modules/cudf/build/js/types/mappings.js:97:24)
    at asColumn (/opt/node-rapids/modules/cudf/build/js/series.js:803:33)
[...]

PR #268, opened 2021-07-08
Create Series constructor for Array and Buffers

love-lena:
Related: #260

### DIFF
--- a/modules/cudf/src/series.ts
+++ b/modules/cudf/src/series.ts
@@ -12,7 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {MemoryData} from '@nvidia/cuda';
+import {
+  Int32Buffer,
+  MemoryData,
+  TypedArray,
+} from '@nvidia/cuda';
 import {DeviceBuffer, MemoryResource} from '@rapidsai/rmm';
 import * as arrow from 'apache-arrow';
 import {VectorType} from 'apache-arrow/interfaces';
@@ -202,6 +206,10 @@ export class AbstractSeries<T extends DataType = any> {
    * ```
    */
   static new<T extends DataType>(input: AbstractSeries<T>|Column<T>|SeriesProps<T>): Series<T>;
+
+  static new<T extends DataType>(input: Int32Array): Series<T>;
+
+  static new<T extends DataType>(input: Int32Buffer): Series<T>;
   /**
    * Create a new cudf.StringSeries
    *
@@ -238,6 +246,7 @@ export class AbstractSeries<T extends DataType = any> {
    * ```
    */
   static new(input: (bigint|null|undefined)[]): Series<Int64>;
+
   /**
    * Create a new cudf.Bool8Series
    *
@@ -334,11 +343,12 @@ export class AbstractSeries<T extends DataType = any> {
   static new(input: (Date|null|undefined)[][]): Series<List<TimestampMillisecond>>;
 
   static new<T extends DataType>(input: AbstractSeries<T>|Column<T>|SeriesProps<T>|arrow.Vector<T>|
-                                 (string|null|undefined)[]|(number|null|undefined)[]|
-                                 (bigint|null|undefined)[]|(boolean|null|undefined)[]|
-                                 (Date|null|undefined)[]|(string|null|undefined)[][]|
-                                 (number|null|undefined)[][]|(bigint|null|undefined)[][]|
-                                 (boolean|null|undefined)[][]|(Date|null|undefined)[][]) {
+                                 (string|null|undefined)[]|Int32Array|Int32Buffer|
+                                 (number|null|undefined)[]|(bigint|null|undefined)[]|
+                                 (boolean|null|undefined)[]|(Date|null|undefined)[]|
+                                 (string|null|undefined)[][]|(number|null|undefined)[][]|
+                                 (bigint|null|undefined)[][]|(boolean|null|undefined)[][]|
+                                 (Date|null|undefined)[][]) {
     return columnToSeries(asColumn<T>(input)) as any as Series<T>;
   }
 
@@ -1207,7 +1217,7 @@ export {
   StructSeries,
 };
 
-function inferType(value: any[]): DataType {
+function inferType(value: any[]|TypedArray): DataType {
   if (value.length === 0) { return new Float64; }
   let nullsCount    = 0;
   let arraysCount   = 0;
@@ -1272,6 +1282,7 @@ function inferType(value: any[]): DataType {
 function asColumn<T extends DataType>(
   value: AbstractSeries<T>|SeriesProps<T>|Column<T>|arrow.Vector<T>  //
   |(string | null | undefined)[]                                     //
+  |Int32Array|Int32Buffer                                            //
   |(number | null | undefined)[]                                     //
   |(bigint | null | undefined)[]                                     //
   |(boolean | null | undefined)[]                                    //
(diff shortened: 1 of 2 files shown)


love-lena:
I have added the Series constructors for Int32Arrays and Int32Buffers, and it works. Now I want to generalize it, and I am not sure of the best way to do that.

My list of things it should work for are:
- Int8Array
- Int16Array
- Int32Array
- Int64Array
- more

modules/cudf/src/series.ts
```diff
@@ -202,6 +206,10 @@ export class AbstractSeries<T extends DataType = any> {
    * ```
    */
   static new<T extends DataType>(input: AbstractSeries<T>|Column<T>|SeriesProps<T>): Series<T>;
+
+  static new<T extends DataType>(input: Int32Array): Series<T>;
```
trxcllnt, line 210:
```suggestion
  static new(input: Int32Array): Series<Int32>;
```

modules/cudf/src/series.ts
```diff
@@ -202,6 +206,10 @@ export class AbstractSeries<T extends DataType = any> {
...
+
+  static new<T extends DataType>(input: Int32Buffer): Series<T>;
```
trxcllnt, line 212:
```suggestion
  static new(input: Int32Buffer): Series<Int32>;
```

modules/cudf/src/series.ts
```diff
@@ -1287,6 +1298,15 @@ function asColumn<T extends DataType>(
...
              {type: inferType(value), values: value as any, highWaterMark: Infinity})) as any;
   }
+  if (value instanceof Int32Array) {
+    return fromArrow(arrow.Vector.from(
+             {type: new Int32, values: value as any, highWaterMark: Infinity})) as any;
+  }
+  if (value instanceof Int32Buffer) {
+    const value_arr = new value.TypedArray(value);
+    return fromArrow(arrow.Vector.from(
+             {type: new Int32, values: value_arr as any, highWaterMark: Infinity})) as any;
```
trxcllnt, line 1308:
```suggestion
    return new Column({ type: new Int32, data: value, length: value.length });
```

modules/cudf/src/series.ts
```diff
@@ -1287,6 +1298,15 @@ function asColumn<T extends DataType>(
     return fromArrow(arrow.Vector.from(
              {type: inferType(value), values: value as any, highWaterMark: Infinity})) as any;
   }
+  if (value instanceof Int32Array) {
+    return fromArrow(arrow.Vector.from(
+             {type: new Int32, values: value as any, highWaterMark: Infinity})) as any;
```
trxcllnt, line 1303:
```suggestion
    return new Column({ type: new Int32, data: value, length: value.length });
```

trxcllnt (changes requested):
Left a few suggestions. Can simplify by returning a concrete Series since the inputs are host/device arrays of known byte-widths.

bryevdv:
@trxcllnt would it be possible to make a generic function that extracts the input buffer/array type and maps it to the dtype, in order to collapse the number of explicit signatures and codepaths when more buffer/array types are added? I'm thinking similar to `ArrowToCUDFType` but perhaps thats not applicable in the same way?

trxcllnt:
@bryevdv it is definitely possible, but from my experience trying similar things in Arrow, likely more trouble than it's worth

love-lena:
I will work on a version thats just very explicit with the data then. For the 8-32 bit int types it wont be too much, if there are more Array/Buffer types we want to add overloads for it will get a little messier

modules/cudf/src/series.ts
```diff
@@ -237,7 +268,9 @@ export class AbstractSeries<T extends DataType = any> {
    * const a = Series.new([1n, 2n, 3n, undefined, 4n]);
    * ```
    */
-  static new(input: (bigint|null|undefined)[]): Series<Int64>;
+  static new(input: (bigint|null|undefined)[]|BigInt64Array|Int64Buffer): Series<Int64>;
+  // static new(input: Float64Array|Float64Buffer): Series<Float64>;
```
love-lena, line 272: delete

modules/cudf/src/series.ts
```diff
@@ -1207,7 +1242,7 @@ export {
   StructSeries,
 };
 
-function inferType(value: any[]): DataType {
+function inferType(value: any[]|TypedArray): DataType {
```
love-lena, line 1245: nope

modules/cudf/src/series.ts
```diff
@@ -1269,39 +1304,87 @@ function inferType(value: any[]): DataType {
...
+//     case BigInt64Array: return new Int64;
+//     case Uint8Array: return new Uint8;
+//     case Uint16Array: return new Uint16;
+//     case Uint32Array: return new Uint32;
+//     case BigUint64Array: return new Uint64;
+//     case Float32Array: return new Float32;
+//     case Float64Array: return new Float64;
+//     default: break;
+//   }
+// }
```
love-lena, line 1321: why me

modules/cudf/src/series.ts
```diff
@@ -1269,39 +1304,87 @@ function inferType(value: any[]): DataType {
...
+    return new Column({type: new Uint16, data: value, length: value.length});
+  } else if (value instanceof Uint32Array || value instanceof Uint32Buffer) {
+    return new Column({type: new Uint32, data: value, length: value.length});
+  } else if (value instanceof BigUint64Array || value instanceof Uint64Buffer) {
+    return new Column({type: new Uint64, data: value, length: value.length});
+  } else if (value instanceof Float32Array || value instanceof Float32Buffer) {
+    return new Column({type: new Float32, data: value, length: value.length});
+  } else if (value instanceof Float64Array || value instanceof Float64Buffer) {
+    return new Column({type: new Float64, data: value, length: value.length});
+  }
```
love-lena, line 1377: _|___|___|___|___|___|___|___|___|___|___|___|
___|___|___|___|___|___|___|___|___|___|___|_
_|___|___|___|___|___|___|___|___|___|___|___|
___|___|___|___|___|___|___|___|___|___|___|_
_|___|___|___|___|___|___|___|___|___|___|___|
___|___|___|___|___|___|___|___|___|___|___|_
_|___|___|___|___|___|___|___|___|___|___|___

love-lena, line 1377:
```
_|___|___|___|___|___|___|___|___|___|___|___|
___|___|___|___|___|___|___|___|___|___|___|_
_|___|___|___|___|___|___|___|___|___|___|___|
___|___|___|___|___|___|___|___|___|___|___|_
_|___|___|___|___|___|___|___|___|___|___|___|
___|___|___|___|___|___|___|___|___|___|___|_
_|___|___|___|___|___|___|___|___|___|___|___
```

trxcllnt: Honestly this is fine with me. If you prefer, you can make it a switch-case (like below). Neither is particularly pleasing, but like I mentioned earlier today, probably not worth agonizing over.

```suggestion
  switch (value.constructor) {
    case Int8Buffer:
    case Int8Array: return new Column({type: new Int8, data: value});
    case Int16Buffer:
    case Int16Array: return new Column({type: new Int16, data: value});
    case Int32Buffer:
    case Int32Array: return new Column({type: new Int32, data: value});
    case Int64Buffer:
    case BigInt64Array: return new Column({type: new Int64, data: value});
    case Uint8Buffer:
    case Uint8Array: return new Column({type: new Uint8, data: value});
    case Uint16Buffer:
    case Uint16Array: return new Column({type: new Uint16, data: value});
    case Uint32Buffer:
    case Uint32Array: return new Column({type: new Uint32, data: value});
    case Uint64Buffer:
    case BigUint64Array: return new Column({type: new Uint64, data: value});
    case Float32Buffer:
    case Float32Array: return new Column({type: new Float32, data: value});
    case Float64Buffer:
    case Float64Array: return new Column({type: new Float64, data: value});
    default: break;
  }
```

trxcllnt (approved):

PR #270, opened 2021-07-12
[FEA] Make blazingsql debug builds work, add `BlazingContext.sql` implementation

trxcllnt:
* Fixes issues blocking blazingsql debug builds
* Adds an initial single-node `BlazingContext.sql` implementation


mzegar:
I forced pushed a wrong branch or something... pushed this branch since it seems to have been what I wanted to do https://github.com/rapidsai/node-rapids/pull/271

PR #271, opened 2021-07-12
[FEA] Make blazingsql debug builds work, add BlazingContext.sql implementation 

mzegar:
- Fixes issues blocking blazingsql debug builds
- Adds an initial single-node BlazingContext.sql implementation


mzegar:
Received permission from Paul to ship it.

PR #273, opened 2021-07-13
Refactor cast to use Visitor pattern

### LINKED ISSUE #262: Refactor the cast() implementation so different subtypes can override
Suggestion from @trxcllnt 

> because there's sometimes custom conversion logic between different types (like string and int)

#### Proposed change

```
class CastVisitor extends Arrow.Visitor {
  constructor(private series) { super(); }
  visitInt8(dtype) { return this.series._castAsInt8(); }
  visitInt16(dtype) { return this.series._castAsInt16(); }
  visitInt32(dtype) { return this.series._castAsInt32(); }
  // etc
}

class Series {
  cast(dtype) {
    return new CastVisitor(this).visit(dtype);
  }
  protected _castAsInt8(): Series<Int8> { throw new Error("Unimplemented"); }
  protected _castAsInt16(): Series<Int16> { throw new Error("Unimplemented"); }
  protected _castAsInt32(): Series<Int32> { throw new Error("Unimplemented"); }
  // etc
}

class StringSeries {
  protected _castAsInt8() { return Series.new(this._col.stoi(new Int8)); }
  protected _castAsInt16() { return Series.new(this._col.stoi(new Int16)); }
  protected _castAsInt32() { return Series.new(this._col.stoi(new Int16)); }
  // etc
}
```

bryevdv:
closes #262 

Still needs a little work:

* [x] add new string-to-numeric casts (with tests)
* [x] add missing cast tests for timestamps

as well as any other casts that I have missed that come up in review

### DIFF
--- a/modules/cudf/src/column.cpp
+++ b/modules/cudf/src/column.cpp
@@ -140,12 +140,12 @@ Napi::Function Column::Init(Napi::Env const& env, Napi::Object exports) {
                        InstanceMethod<&Column::count_re>("countRe"),
                        InstanceMethod<&Column::matches_re>("matchesRe"),
                        // column/convert.cpp
-                       InstanceMethod<&Column::is_float>("isFloat"),
-                       InstanceMethod<&Column::from_floats>("stringsFromFloats"),
-                       InstanceMethod<&Column::to_floats>("stringsToFloats"),
-                       InstanceMethod<&Column::is_integer>("isInteger"),
-                       InstanceMethod<&Column::from_integers>("stringsFromIntegers"),
-                       InstanceMethod<&Column::to_integers>("stringsToIntegers"),
+                       InstanceMethod<&Column::string_is_float>("stringIsFloat"),
+                       InstanceMethod<&Column::strings_from_floats>("stringsFromFloats"),
+                       InstanceMethod<&Column::strings_to_floats>("stringsToFloats"),
+                       InstanceMethod<&Column::string_is_integer>("stringIsInteger"),
+                       InstanceMethod<&Column::strings_from_integers>("stringsFromIntegers"),
+                       InstanceMethod<&Column::strings_to_integers>("stringsToIntegers"),
                      });
 }
 
--- a/modules/cudf/src/column.cpp
+++ b/modules/cudf/src/column.cpp
@@ -141,11 +141,11 @@ Napi::Function Column::Init(Napi::Env const& env, Napi::Object exports) {
                        InstanceMethod<&Column::matches_re>("matchesRe"),
                        // column/convert.cpp
                        InstanceMethod<&Column::is_float>("isFloat"),
-                       InstanceMethod<&Column::from_floats>("fromFloats"),
-                       InstanceMethod<&Column::to_floats>("toFloats"),
+                       InstanceMethod<&Column::from_floats>("stringsFromFloats"),
+                       InstanceMethod<&Column::to_floats>("stringsToFloats"),
                        InstanceMethod<&Column::is_integer>("isInteger"),
-                       InstanceMethod<&Column::from_integers>("fromIntegers"),
-                       InstanceMethod<&Column::to_integers>("toIntegers"),
+                       InstanceMethod<&Column::from_integers>("stringsFromIntegers"),
+                       InstanceMethod<&Column::to_integers>("stringsToIntegers"),
                      });
 }
 
--- a/modules/cudf/src/column.cpp
+++ b/modules/cudf/src/column.cpp
@@ -139,6 +139,13 @@ Napi::Function Column::Init(Napi::Env const& env, Napi::Object exports) {
                        InstanceMethod<&Column::contains_re>("containsRe"),
                        InstanceMethod<&Column::count_re>("countRe"),
                        InstanceMethod<&Column::matches_re>("matchesRe"),
+                       // column/convert.cpp
+                       InstanceMethod<&Column::is_float>("isFloat"),
+                       InstanceMethod<&Column::from_floats>("fromFloats"),
+                       InstanceMethod<&Column::to_floats>("toFloats"),
+                       InstanceMethod<&Column::is_integer>("isInteger"),
+                       InstanceMethod<&Column::from_integers>("fromIntegers"),
+                       InstanceMethod<&Column::to_integers>("toIntegers"),
                      });
 }
 
--- a/modules/cudf/src/column.ts
+++ b/modules/cudf/src/column.ts
@@ -726,7 +726,7 @@ export interface Column<T extends DataType = any> {
    *
    *  @returns A string Column with floats as strings.
    */
-  fromFloats(memoryResource?: MemoryResource): Column<Utf8String>;
+  stringsFromFloats(memoryResource?: MemoryResource): Column<Utf8String>;
 
   /**
    * Returns a new floating point numeric column parsing float values from the provided
@@ -743,7 +743,7 @@ export interface Column<T extends DataType = any> {
    *
    *  @returns A Column of a the specified float type with the results of the conversion.
    */
-  toFloats<R extends DataType>(dataType: R, memoryResource?: MemoryResource): Column<R>;
+  stringsToFloats<R extends DataType>(dataType: R, memoryResource?: MemoryResource): Column<R>;
 
   /**
    * Creates a column of `BOOL8` elements indicating strings in which all characters are valid for
@@ -770,7 +770,7 @@ export interface Column<T extends DataType = any> {
    *
    *  @returns A string Column with integers as strings.
    */
-  fromIntegers(memoryResource?: MemoryResource): Column<Utf8String>;
+  stringsFromIntegers(memoryResource?: MemoryResource): Column<Utf8String>;
 
   /**
    * Returns a new integer numeric column parsing integer values from the provided strings column.
@@ -791,7 +791,7 @@ export interface Column<T extends DataType = any> {
    *
    *  @returns A Column of a the specified integral type with the results of the conversion.
    */
-  toIntegers<R extends DataType>(dataType: R, memoryResource?: MemoryResource): Column<R>;
+  stringsToIntegers<R extends DataType>(dataType: R, memoryResource?: MemoryResource): Column<R>;
 
   /**
    * Compute the trigonometric sine for each value in this Column.
--- a/modules/cudf/src/column.ts
+++ b/modules/cudf/src/column.ts
@@ -26,6 +26,7 @@ import {
   Int64,
   Integral,
   Numeric,
+  Utf8String,
 } from './types/dtypes';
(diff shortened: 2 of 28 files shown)


modules/cudf/src/series.ts
```diff
@@ -408,36 +479,38 @@ export class AbstractSeries<T extends DataType = any> {
...
+    // 1. Join this Series' values with the `categories` Series to determine the index
+    // positions
     //    (i.e. `codes`) of the values to keep.
     // 2. Sort the codes by the original value's position in this Series.
     // 3. Replace missing codes with `nullSentinel`.
     //
-    // Note: Written as a single expression so the intermediate memory allocated for the `join` and
-    // `sortValues` calls are GC'd as soon as possible.
+    // Note: Written as a single expression so the intermediate memory allocated for the
+    // `join` and `sortValues` calls are GC'd as soon as possible.
```
bryevdv, line 513: would be good to figure out what about my config is somehow different and causing these spurious changes

mzegar: I'm pretty sure this has happened to me as well.

modules/cudf/src/column.ts
```diff
@@ -700,6 +701,98 @@ export interface Column<T extends DataType = any> {
    */
   isNotNaN(memoryResource?: MemoryResource): Column<Bool8>;
 
+  /**
+   * Creates a column of `BOOL8` elements indicating strings in which all characters are valid for
```
mzegar, line 705: Maybe I'm misunderstanding, but, `indicating strings`?

modules/cudf/src/column.ts
```diff
@@ -700,6 +701,98 @@ export interface Column<T extends DataType = any> {
...
+   * For each float, a string is created in base-10 decimal. Negative numbers will include a '-'
+   * prefix. Numbers producing more than 10 significant digits will produce a string that includes
+   * scientific notation (e.g. "-1.78e+15").
+   *
+   * @param memoryResource The optional MemoryResource used to allocate the result Column's device
+   *   memory.
+   *
+   *  @returns A string Column with floats as strings.
+   */
+  fromFloats(memoryResource?: MemoryResource): Column<Utf8String>;
```
mzegar, line 729: Should this method be inside of `FloatSeries`?

modules/cudf/src/column.ts
```diff
@@ -700,6 +701,98 @@ export interface Column<T extends DataType = any> {
...
+   * @param dataType Type of floating numeric column to return.
+   * @param memoryResource  The optional MemoryResource used to allocate the result Column's device
+   *   memory.
+   *
+   *  @returns A Column of a the specified float type with the results of the conversion.
+   */
+  toFloats<R extends DataType>(dataType: R, memoryResource?: MemoryResource): Column<R>;
+
+  /**
+   * Creates a column of `BOOL8` elements indicating strings in which all characters are valid for
```
mzegar, line 749: Same comment as above.

modules/cudf/src/column.ts
```diff
@@ -700,6 +701,98 @@ export interface Column<T extends DataType = any> {
...
+   *
+   * For each integer, a string is created in base-10 decimal. Negative numbers will include a '-'
+   * prefix.
+   *
+   * @param memoryResource The optional MemoryResource used to allocate the result Column's device
+   *   memory.
+   *
+   *  @returns A string Column with integers as strings.
+   */
+  fromIntegers(memoryResource?: MemoryResource): Column<Utf8String>;
```
mzegar, line 773: Same comment as above.

modules/cudf/src/node_cudf/column.hpp
```diff
@@ -722,6 +722,28 @@ struct Column : public EnvLocalObjectWrap<Column> {
     std::string const& pattern,
     rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource()) const;
 
+  // column/convert.cpp
+
```
mzegar, line 726: Extreme nit pick, typically we'd remove this new line.

mzegar (commented):
Some questions / comments

modules/cudf/src/column.ts
bryevdv: I tried to keep the verbiage similar to out other usage, but the libcudf doc this was cribbed from reads
> Returns a boolean column identifying strings in which all characters are valid for conversion to floats.

Do you think that "identifying" would be more clear than "indicating"? These are basically synonymous usage in my mind but also happy to change

modules/cudf/src/column.ts
bryevdv: No these are all methods that go on the low-level `Column`, because they call out to `libcudf` directly. This is similar to the way `containsRe`  is on `Column` even though it only applies specifically to *string* columns. I agree that the wide API on column is not perfect in some ways, but it seems to be an acceptable compromise given that it is not user-facing.

modules/cudf/src/node_cudf/column.hpp
bryevdv: not sure where that came from!

modules/cudf/src/column.ts
mzegar: Chatted with Bryan about this, this is fine.

bryevdv: We seem to use "indicating" pretty consistently throughout already in similar bool returns so I'll keep this for now to be consistent but maybe we can/should change across the board.

modules/cudf/src/column.ts
```diff
@@ -700,6 +701,98 @@ export interface Column<T extends DataType = any> {
...
+   * Only characters [0-9] plus a prefix '-' and '+' and decimal '.' are recognized. Additionally,
+   * scientific notation is also supported (e.g. "-1.78e+5").
+   *
+   * @param dataType Type of floating numeric column to return.
+   * @param memoryResource  The optional MemoryResource used to allocate the result Column's device
+   *   memory.
+   *
+   *  @returns A Column of a the specified float type with the results of the conversion.
+   */
+  toFloats<R extends DataType>(dataType: R, memoryResource?: MemoryResource): Column<R>;
```
trxcllnt, line 746: only suggestion w/ this and `fromFloats` is maybe update the name to indicate it's for strings? This is in the `cudf::strings` namespace so the behavior is clear, but here it seems like we could use it for any dtype?

```suggestion
  stringsToFloats<R extends DataType>(dataType: R, memoryResource?: MemoryResource): Column<R>;
```

mzegar: ah yeah, I spoke with Bryan about my confusion that this was for strings, I think this is probably good solution.

trxcllnt (commented):
Looking great so far, just one comment about the naming of string-specific methods.

modules/cudf/src/column.ts
bryevdv: Seems reasonable, I was just hewing the the existing libcudf names as a default

modules/cudf/src/column.ts
```diff
@@ -700,6 +701,98 @@ export interface Column<T extends DataType = any> {
...
+   * Only characters [0-9] plus a prefix '-' and '+' and decimal '.' are recognized. Additionally,
+   * scientific notation is also supported (e.g. "-1.78e+5").
+   *
+   * @param dataType Type of floating numeric column to return.
+   * @param memoryResource  The optional MemoryResource used to allocate the result Column's device
+   *   memory.
+   *
+   *  @returns A Column of a the specified float type with the results of the conversion.
+   */
+  stringsToFloats<R extends DataType>(dataType: R, memoryResource?: MemoryResource): Column<R>;
```
trxcllnt, line 746:
```suggestion
  stringsToFloats<R extends Float32 | Float64>(dataType: R, memoryResource?: MemoryResource): Column<R>;
```

modules/cudf/src/column.ts
```diff
@@ -700,6 +701,98 @@ export interface Column<T extends DataType = any> {
...
+  /**
+   * Creates a column of `BOOL8` elements indicating strings in which all characters are valid for
+   * conversion to floats.
+   *
+   * @param memoryResource The optional MemoryResource used to allocate the result Column's device
+   *   memory.
+   * @returns A non-nullable column of `BOOL8` elements with `true` representing convertible
+   *   values
+   */
+  isFloat(memoryResource?: MemoryResource): Column<Bool8>;
```
trxcllnt, line 713: String naming should be here too, since we can't call this on other dtypes right?

```suggestion
  stringIsFloat(memoryResource?: MemoryResource): Column<Bool8>;
```

modules/cudf/src/column.cpp
```diff
@@ -139,6 +139,13 @@ Napi::Function Column::Init(Napi::Env const& env, Napi::Object exports) {
                        InstanceMethod<&Column::contains_re>("containsRe"),
                        InstanceMethod<&Column::count_re>("countRe"),
                        InstanceMethod<&Column::matches_re>("matchesRe"),
+                       // column/convert.cpp
+                       InstanceMethod<&Column::is_float>("isFloat"),
+                       InstanceMethod<&Column::from_floats>("stringsFromFloats"),
+                       InstanceMethod<&Column::to_floats>("stringsToFloats"),
+                       InstanceMethod<&Column::is_integer>("isInteger"),
+                       InstanceMethod<&Column::from_integers>("stringsFromIntegers"),
+                       InstanceMethod<&Column::to_integers>("stringsToIntegers"),
```
trxcllnt, line 148: Rename on the C++ side as well?

```suggestion
                       InstanceMethod<&Column::string_is_float>("stringIsFloat"),
                       InstanceMethod<&Column::strings_from_floats>("stringsFromFloats"),
                       InstanceMethod<&Column::strings_to_floats>("stringsToFloats"),
                       InstanceMethod<&Column::string_is_integer>("stringIsInteger"),
                       InstanceMethod<&Column::strings_from_integers>("stringsFromIntegers"),
                       InstanceMethod<&Column::to_integers>("stringsToIntegers"),
```

modules/cudf/src/column.ts
```diff
@@ -700,6 +701,98 @@ export interface Column<T extends DataType = any> {
...
+  /**
+   * Creates a column of `BOOL8` elements indicating strings in which all characters are valid for
+   * conversion to floats.
+   *
+   * @param memoryResource The optional MemoryResource used to allocate the result Column's device
+   *   memory.
+   * @returns A non-nullable column of `BOOL8` elements with `true` representing convertible
+   *   values
+   */
+  isInteger(memoryResource?: MemoryResource): Column<Bool8>;
```
trxcllnt, line 757:
```suggestion
  stringIsInteger(memoryResource?: MemoryResource): Column<Bool8>;
```

bryevdv:
@trxcllnt I think this is more or less ready however I would describe the timestamp tests as "plausible but suspect" so attention there is appreciated

trxcllnt (approved):

PR #275, opened 2021-07-15
[FEA] bsql managing tables bindings

mzegar:
https://docs.blazingsql.com/reference/python/blazing_context.html#managing-tables

partition is deprecated, so didn't add.


mzegar:
Going to be a draft for now, going to implement some more of these `managing` table methods.

trxcllnt (approved):

Issue #259, opened 2021-07-02
Issues with empty DataFrames


bryevdv:
There are two cases that lead small issues relate to empty DataFrames:

#####  No arguments to `DataFrame`
```
> {DataFrame} = require("@rapidsai/cudf")

> const df = new DataFrame()
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at _seriesToColumns (/opt/node-rapids/modules/cudf/build/js/data_frame.js:31:41)
    at new DataFrame (/opt/node-rapids/modules/cudf/build/js/data_frame.js:95:110)
```
Presumably this should function identically to `new DataFrame({})`, which does work

##### `numRows` on empty `DataFrame`
```
> {DataFrame} = require("@rapidsai/cudf")

> new DataFrame({}).numRows 
Uncaught TypeError: Cannot read property 'length' of undefined
    at DataFrame.get numRows [as numRows] (/opt/node-rapids/modules/cudf/build/js/data_frame.js:112:54)
```
should return 0

PR #290, opened 2021-08-09
Fea: Node SQL module + Node SQL Server Demo

mzegar:
- Adds SQL module
   - SQL  is a module that spins up multiple SQLContext instances
- Adds SQL server demo
   - Uses sql builder to build a basic query, calls into a SQLContext instance

Screenshot from 2021-09-22 13-06-52


modules/blazingsql/src/blazingcluster.ts
```diff
@@ -0,0 +1,114 @@
...
+interface BlazingCusterProps {
+  numWorkers: number;
+}
+
+export class BlazingCluster {
+  workers: ChildProcess[];
+  bc: BlazingContext;
+
+  constructor({numWorkers = 1}: BlazingCusterProps) {
+    this.workers = Array(numWorkers).fill(fork(`${__dirname}/worker`, {serialization: 'advanced'}));
```
trxcllnt, line 38: won't this fill all the slots of the `workers` Array with the same ChildProcess instance?

modules/blazingsql/src/blazingcluster.ts
```diff
@@ -0,0 +1,306 @@
...
+          } = msg;
+
+          if (operation === QUERY_RAN) {
+            resolve({ctxToken, messageId, df: this.blazingContext.pullFromCache(messageId)});
+          }
+        });
+      }));
+    });
+
+    let result_df = new DataFrame({a: Series.new([])});
```
trxcllnt, line 289:
```suggestion
    let result_df = new DataFrame({});
```

modules/blazingsql/src/blazingcluster.ts
```diff
@@ -0,0 +1,306 @@
...
+
+    await Promise.all(queryPromises).then(function(results) {
+      results.forEach((result: any) => {
+        const {df}: {df: DataFrame} = result;
```
trxcllnt, line 293:
```suggestion
      results.forEach(({df}: {df: DataFrame}) => {
```

modules/blazingsql/src/blazingcluster.ts
```diff
@@ -0,0 +1,306 @@
...
+   * @param numWorkers the number of child processes to spawn
+   *
+   * @example
+   * ```typescript
+   * import {BlazingCluster} from '@rapidsai/blazingsql';
+   *
+   * await BlazingContext.init();
+   * ```
+   */
+  static async init(numWorkers = 1): Promise<BlazingCluster> {
```
trxcllnt, line 57: Will need to expose some more config options for users too -- ip address, ports list (or port range), network interface etc.

mzegar: yeah agreed, I think we should allow people to setup logging as you did when debugging.

modules/blazingsql/src/blazingcluster.ts
```diff
@@ -0,0 +1,306 @@
...
+    });
+
+    return result_df;
+  }
+
+  stop(): void {
+    this.workers.forEach((worker) => { worker.kill('SIGKILL'); });
+  }
+
+  private generateMessageId(ctxToken: number): string { return `message_${ctxToken}`; }
```
mzegar, line 305: I think this will show up on the object prototype, so going to make this an ignored function at the top of the file.

modules/blazingsql/src/blazingcluster.ts
mzegar: https://github.com/rapidsai/node/issues/259
We need to keep it as is due to https://github.com/rapidsai/node/issues/259

trxcllnt: Why not just include the fix in this PR?

mzegar: No good reason, I guess the idea was to keep PRs pretty separate. I'll merge and include the fixes in this branch though.

modules/blazingsql/src/blazingcluster.ts
```diff
@@ -0,0 +1,320 @@
...
+    await Promise.all(createContextPromises);
+
+    return bc;
+  }
+
+  private constructor(numWorkers: number) {
+    // If `__dirname` includes '/src' we are currently running a Jest test. Use a different relative
+    // path for when we are in a Jest test versus normal usage.
+    const relativePath = __dirname.includes('/src') ? './build/js/worker.js'
+                                                    : './modules/blazingsql/build/js/worker.js';
```
trxcllnt, line 108: We need to resolve this path from whatever `__dirname` is in this file. `modules/blazingsql` will not resolve to anything when this code is installed in a `node_modules` directory in the runtime containers.

mzegar: Should be fixed.

package.json
```diff
@@ -58,14 +58,16 @@
...
+    "modules/demo/client-server",
+    "modules/demo/blazingcluster-server"
   ],
   "dependencies": {
     "@typescript-eslint/eslint-plugin": "4.26.0",
     "@typescript-eslint/parser": "4.26.0",
     "cmake-js": "6.0.0",
     "dotenv": "8.2.0",
     "eslint": "7.27.0",
+    "fastify": "3.20.2",
```
trxcllnt, line 70: do we need this here?

modules/demo/blazingcluster-server/index.js
```diff
@@ -0,0 +1,66 @@
...
+          return result;
+        },
+        ''),
+      resultsCount: df.numRows,
+      queryTime: queryTime
+    });
+  }
+
+  reply.send({result: 'Failed to parse query.'})
+  })
```
trxcllnt, line 60:
```suggestion
    reply.send(new Error('Failed to parse query.'));
  });
```

modules/demo/blazingcluster-server/index.js
```diff
@@ -0,0 +1,66 @@
...
+  async () => {
+  bc = await BlazingCluster.init(2);
+  await bc.createTable('test_table', DataFrame.readCSV({
+    header: 0,
+    sourceType: 'files',
+    sources: [`${__dirname}/small.csv`],
+  }));
+}
+
+init();
```
trxcllnt, line 33: Should wrap this in a call to `fastify.register()` so it blocks server startup.

modules/demo/blazingcluster-server/components/querydashboard.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  onQueryChange = (updatedQuery) => {
+    this.setState({ query: updatedQuery, queryButtonEnabled: updatedQuery.length });
+  }
+
+  async runQuery() {
+    if (this.state.queryButtonEnabled) {
+      this.setState({queryButtonEnabled: false});
+      await fetch(`http://localhost:3000/run_query?sql=${this.state.query}`)
```
trxcllnt, line 39:
```suggestion
      await fetch(`/run_query?sql=${this.state.query}`)
```

lerna.json
```diff
@@ -13,6 +13,7 @@
     "modules/demo/ipc/*",
     "modules/demo/ssr/*",
     "modules/demo/tfjs/*",
-    "modules/demo/client-server"
+    "modules/demo/client-server",
+    "modules/demo/blazingcluster-server"
```
trxcllnt, line 17: slight nit, but what about making a `modules/demo/blazingsql` directory, so that way we have a place to put more BSQL demos?

mzegar: sure I'm assuming similar to `deck` or like `ipc`

modules/blazingsql/src/execution_graph.ts
```diff
@@ -16,21 +16,22 @@ import {DataFrame, Series} from '@rapidsai/cudf';
 import {ExecutionGraphWrapper} from './node_blazingsql';
 
 export class ExecutionGraph {
-  private executionGraphWrapper: ExecutionGraphWrapper;
+  private executionGraphWrapper: ExecutionGraphWrapper|undefined;
 
-  constructor(executionGraphWrapper: ExecutionGraphWrapper) {
+  constructor(executionGraphWrapper?: ExecutionGraphWrapper) {
```
mzegar, line 21: We need some notion of an empty `ExecutionGraph` since we now return early here
https://github.com/rapidsai/node/blob/a54e601eb07eeeecf27c7635e7d3d6b0401af9c4/modules/blazingsql/src/blazingcontext.ts#L221

Any thoughts on representing an empty `ExecutionGraph`?

trxcllnt: We could have an `EmptyExecutionGraph` class?

mzegar: yeah simple enough, thanks.

PR #295, opened 2021-08-23
[FIX] Empty dataframe initialization

### LINKED ISSUE #259: Issues with empty DataFrames

mzegar:
I needed this fix for my current branch which uses an empty `dataframe`. I did a search for any other `columns[0]` accessing without checking length but couldn't find any others.

Closes #259

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -142,7 +142,7 @@ export class DataFrame<T extends TypeMap = any> {
    *
    * ```
    */
-  constructor(data: ColumnAccessor<T>|SeriesMap<T>) {
+  constructor(data: ColumnAccessor<T>|SeriesMap<T> = {} as SeriesMap<T>) {
     this._accessor =
       (data instanceof ColumnAccessor) ? data : new ColumnAccessor(_seriesToColumns(data));
   }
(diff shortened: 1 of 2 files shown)


modules/cudf/src/data_frame.ts
```diff
@@ -162,7 +162,7 @@ export class DataFrame<T extends TypeMap = any> {
    * df.numRows // 2
    * ```
    */
-  get numRows() { return this._accessor.columns[0].length; }
+  get numRows() { return this._accessor.columns.length > 0 ? this._accessor.columns[0].length : 0; }
```
mzegar, line 165: Was also considering...
```js
return this._accessor.columns[0]?.length ?? 0;
```
But I think this is a bit more readable.

bryevdv (approved):

bryevdv (approved):

PR #297, opened 2021-09-10
Add pretty-printing df.toString

bryevdv:
Adds basic pretty-printing method to `DataFrame` e.g.

```
> console.log(new DataFrame({'a': a, 'b': b, 'c': c, 'd': d}).toString({maxColWidth: 70}))
  a          b               d
0.0        0.0 ...        -Inf
1.0        1.1 ...         0.0
2.0        2.2 ... 0.693147181
3.0        3.3 ... 1.098612289
4.0        4.4 ... 1.386294361
5.0        5.5 ... 1.609437912
6.0 6.61234567 ... 1.791759469
```

Supports options:
* `maxColumns`
* `maxColWidth`
* `maxRows`
* `width`

Lots of corner cases and "fix one test break another" but I think this is at a good starting point.

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -22,6 +22,7 @@ import {Column} from './column';
 import {ColumnAccessor} from './column_accessor';
 import {concat as concatDataFrames} from './dataframe/concat';
 import {Join, JoinResult} from './dataframe/join';
+import {DataFrameFormatter, DisplayOptions} from './dataframe/print';
 import {GroupByMultiple, GroupByMultipleProps, GroupBySingle, GroupBySingleProps} from './groupby';
 import {Scalar} from './scalar';
 import {AbstractSeries, Series} from './series';
@@ -222,6 +223,16 @@ export class DataFrame<T extends TypeMap = any> {
 
   /** @ignore */
   asTable() { return new Table({columns: this._accessor.columns}); }
+
+  /**
+   *
+   * @param options
+   * @returns void
+   */
+  toString(options?: DisplayOptions): string {
+    const formatter = new DataFrameFormatter(options ?? {}, this);
+    return formatter.render();
+  }
 
   /**
    * Return a new DataFrame containing only specified columns.
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -225,12 +225,14 @@ export class DataFrame<T extends TypeMap = any> {
   asTable() { return new Table({columns: this._accessor.columns}); }
 
   /**
+   * Return a string with a tabular representation of the DataFrame, pretty-printed according to the
+   * options given.
    *
    * @param options
    * @returns void
    */
-  toString(options?: DisplayOptions): string {
-    const formatter = new DataFrameFormatter(options ?? {}, this);
+  toString(options: DisplayOptions = {}): string {
+    const formatter = new DataFrameFormatter(options, this);
     return formatter.render();
   }
 
--- a/modules/cudf/src/dataframe/print.ts
+++ b/modules/cudf/src/dataframe/print.ts
@@ -47,7 +47,7 @@ export type DisplayOptions = {
 }
 
 export class DataFrameFormatter<T extends TypeMap> {
-  private obj: DataFrame;
+  private frame: DataFrame;
   private maxColumns: number;
   private maxColWidth: number;
   private maxRows: number;
@@ -55,7 +55,7 @@ export class DataFrameFormatter<T extends TypeMap> {
   private htrunc: boolean;
   private vtrunc: boolean;
 
-  constructor(options: DisplayOptions, obj: DataFrame<T>) {
+  constructor(options: DisplayOptions, frame: DataFrame<T>) {
     this.maxColumns  = options.maxColumns ?? 20;
     this.maxColWidth = options.maxColWidth ?? 50;
     this.maxRows     = options.maxRows ?? 60;
@@ -63,18 +63,18 @@ export class DataFrameFormatter<T extends TypeMap> {
     this.htrunc      = false;
     this.vtrunc      = false;
 
-    let tmp = this.preprocess(obj);
+    let tmp = this._preprocess(frame);
     while (this._totalWidth(tmp) > this.width) {
       this.htrunc = true;
       const names = [...tmp.names];
       const N     = Math.ceil((names.length - 1) / 2);
       tmp         = tmp.drop([names[N]]);
     }
-    this.obj = tmp;
+    this.frame = tmp;
   }
 
-  preprocess(obj: DataFrame<T>) {
-    let tmp: DataFrame = obj;
+  private _preprocess(frame: DataFrame<T>) {
+    let tmp: DataFrame = frame;
 
     if (this.maxColumns > 0 && tmp.numColumns > this.maxColumns) {
       this.htrunc       = true;
@@ -97,14 +97,14 @@ export class DataFrameFormatter<T extends TypeMap> {
   }
 
   render(): string {
-    const headers: string[] = [...this.obj.names];
-    const widths: number[]  = this._computeWidths(this.obj, headers);
+    const headers: string[] = [...this.frame.names];
+    const widths: number[]  = this._computeWidths(this.frame, headers);
     const lines             = [];
 
     lines.push(this._formatFields(headers, widths, '   '));
 
-    for (let i = 0; i < this.obj.numRows; ++i) {
-      const fields = headers.map((name: string) => this.obj.get(name).getValue(i));
+    for (let i = 0; i < this.frame.numRows; ++i) {
+      const fields = headers.map((name: string) => this.frame.get(name).getValue(i));
       lines.push(this._formatFields(fields, widths));
     }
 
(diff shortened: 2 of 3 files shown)


modules/cudf/src/dataframe/print.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.htrunc      = false;
+    this.vtrunc      = false;
+
+    let tmp = this.preprocess(obj);
+    while (this._totalWidth(tmp) > this.width) {
+      this.htrunc = true;
+      const names = [...tmp.names];
+      const N     = Math.ceil((names.length - 1) / 2);
+      tmp         = tmp.drop([names[N]]);
+    }
```
bryevdv, line 71: This loop could get in to trouble. I see two options:

* bail if new total width ever equals prev total width
* set a min total width and error earlier

modules/cudf/src/data_frame.ts
```diff
@@ -223,6 +224,16 @@ export class DataFrame<T extends TypeMap = any> {
   /** @ignore */
   asTable() { return new Table({columns: this._accessor.columns}); }
 
+  /**
+   *
+   * @param options
+   * @returns void
+   */
+  toString(options?: DisplayOptions): string {
```
mzegar, line 232: Could do

`toString(options: DisplayOptions = {}): string {`

and then avoid the `??`, but it's up to what you prefer.

modules/cudf/src/data_frame.ts
```diff
@@ -223,6 +224,16 @@ export class DataFrame<T extends TypeMap = any> {
   /** @ignore */
   asTable() { return new Table({columns: this._accessor.columns}); }
 
+  /**
+   *
+   * @param options
+   * @returns void
+   */
```
mzegar, line 231: Need some `docstrings` about this right?

modules/cudf/src/dataframe/print.ts
```diff
@@ -0,0 +1,150 @@
...
+
+import {DataFrame} from '../data_frame';
+import {Series} from '../series';
+import {Int32, Utf8String} from '../types/dtypes';
+import {TypeMap} from '../types/mappings';
+
+export type DisplayOptions = {
+  /**
+   * The maximum number of columns to display before truncation. Truncation may also occur
+   * if the total printed width in characters exceeds `widht
```
mzegar, line 23:
```suggestion
   * if the total printed width in characters exceeds `width`
```

modules/cudf/src/dataframe/print.ts
```diff
@@ -0,0 +1,150 @@
...
+   *
+   */
+  maxColumns?: number,
+
+  /**
+   * The maximum width in characters of a column when printing a DataFram. When the column *
```
mzegar, line 29:
```suggestion
   * The maximum width in characters of a column when printing a DataFrame. When the column *
```

modules/cudf/src/dataframe/print.ts
```diff
@@ -0,0 +1,150 @@
...
+import {DataFrame} from '../data_frame';
+import {Series} from '../series';
+import {Int32, Utf8String} from '../types/dtypes';
+import {TypeMap} from '../types/mappings';
+
+export type DisplayOptions = {
+  /**
+   * The maximum number of columns to display before truncation. Truncation may also occur
+   * if the total printed width in characters exceeds `widht
+   *
```
mzegar, line 24:
```suggestion
   * (default: 20)
   *
```

modules/cudf/src/dataframe/print.ts
```diff
@@ -0,0 +1,150 @@
...
+export class DataFrameFormatter<T extends TypeMap> {
+  private obj: DataFrame;
+  private maxColumns: number;
+  private maxColWidth: number;
+  private maxRows: number;
+  private width: number;
+  private htrunc: boolean;
+  private vtrunc: boolean;
+
+  constructor(options: DisplayOptions, obj: DataFrame<T>) {
```
mzegar, line 57: minor nit, maybe `input` or `df` or `dataframe` would be a better name than `obj`? I mention this since the class name is `DataFrameFormatter`

modules/cudf/src/dataframe/print.ts
mzegar: Yeah setting a min value for `width` seems reasonable to me.

modules/cudf/src/dataframe/print.ts
```diff
@@ -0,0 +1,150 @@
...
+    while (this._totalWidth(tmp) > this.width) {
+      this.htrunc = true;
+      const names = [...tmp.names];
+      const N     = Math.ceil((names.length - 1) / 2);
+      tmp         = tmp.drop([names[N]]);
+    }
+    this.obj = tmp;
+  }
+
+  preprocess(obj: DataFrame<T>) {
```
mzegar, line 75: If you decide on changing the `obj` param name above.

mzegar (commented):
Neat stuff, excited to use this myself.

modules/cudf/src/data_frame.ts
bryevdv: https://github.com/rapidsai/node/pull/297/commits/6853b924285a69689aa481378e551aa295b4a947

modules/cudf/src/data_frame.ts
bryevdv: https://github.com/rapidsai/node/pull/297/commits/6853b924285a69689aa481378e551aa295b4a947

modules/cudf/src/dataframe/print.ts
bryevdv: https://github.com/rapidsai/node/pull/297/commits/842a5bd781f3a403a745a759df09936256c70944

modules/cudf/src/dataframe/print.ts
bryevdv: following pandas, I went with `frame` 

https://github.com/rapidsai/node/pull/297/commits/842a5bd781f3a403a745a759df09936256c70944

mzegar (approved):
Looks good on my end, feel free to get some feedback from others as well.

modules/cudf/src/dataframe/print.ts
bryevdv: I will probably add a min `width` check (5 characters?) but there are lots of interactions so I think I will add the clunky `current = prev` check to. But then there's a question of what to do:

* if things get down to zero column / can't get total width < with

Is that an *error*? Or is that "return an empty string"?

mzegar: I'd personally return an empty string, however, is there some convention that the `toString()` methods follow in javascript?

bryevdv: I think I'm inclined that way as well. @trxcllnt do you have a preference here?

modules/cudf/src/dataframe/print.ts
```diff
@@ -0,0 +1,151 @@
...
+  constructor(options: DisplayOptions, frame: DataFrame<T>) {
+    this.maxColumns  = options.maxColumns ?? 20;
+    this.maxColWidth = options.maxColWidth ?? 50;
+    this.maxRows     = options.maxRows ?? 60;
+    this.width       = options.width ?? 80;
+    this.htrunc      = false;
+    this.vtrunc      = false;
+
+    let tmp = this._preprocess(frame);
+    while (this._totalWidth(tmp) > this.width) {
```
trxcllnt, line 67: Does the measured size of the rest of the columns change after dropping a column? Seems like we could measure all the columns once up front, then remove the dropped ones from the corresponding sizes list?

modules/cudf/src/dataframe/print.ts
trxcllnt: We could truncate the last column values if the user requests a pathologically small max width... Maybe our minimum should be 4 -- 1 character of data + an ellipses?

modules/cudf/src/dataframe/print.ts
bryevdv: Sure that seems possible :+1:

modules/cudf/src/dataframe/print.ts
```diff
@@ -0,0 +1,151 @@
...
+  private maxRows: number;
+  private width: number;
+  private htrunc: boolean;
+  private vtrunc: boolean;
+
+  constructor(options: DisplayOptions, frame: DataFrame<T>) {
+    this.maxColumns  = options.maxColumns ?? 20;
+    this.maxColWidth = options.maxColWidth ?? 50;
+    this.maxRows     = options.maxRows ?? 60;
+    this.width       = options.width ?? 80;
```
trxcllnt, line 62: Maybe it's just me, but 80 chars as default max total width seems pretty narrow. Why not default max width to Infinity?

modules/cudf/src/dataframe/print.ts
bryevdv: That's an interesting idea. That is not what pandas does but I can look into it if

modules/cudf/src/dataframe/print.ts
bryevdv: I don't have strong opinions here, this is just following what pandas has for a default. I am fine setting the default to unlimited if that's your preferece

modules/cudf/src/dataframe/print.ts
bryevdv: https://github.com/rapidsai/node/pull/297/commits/510c75317dc52450e67e277ca6d54fdd103f9260

modules/cudf/src/dataframe/print.ts
bryevdv: default to unlimited ("terminal") width in https://github.com/rapidsai/node/pull/297/commits/d481fb773ce5a61125be49216ef9b5e921bc23aa 

note that there is still a default max cols though.

modules/cudf/src/dataframe/print.ts
bryevdv: If our going-in position is default to infinite width, then TBH this seems like a fair amount of complexity for a really rare/unimportant case. I did set a min col width of 4, so there will always be at least some output (even if its just a column of elipses). Any objection if I leave it at that?

trxcllnt: If I'm reading this right, we'll drop all the columns if the max width is smaller than the width of the smallest column? I can live with that.

PR #304, opened 2021-09-30
`GLFW` Demo Titles

mzegar:
Let's give the `GLFW` windows a title instead of being empty or `untitled`.


AjayThorve (approved):
looks good!

PR #307, opened 2021-10-04
DRAFT: SQLCluster allow for CSV files

mzegar:


modules/sql/src/cluster.ts
```diff
@@ -121,10 +121,39 @@ export class SQLCluster {
...
+      await Promise.all(this._workers.map((worker, i) => worker.createTable(tableName, ids[i])));
+    } else {
+      const chunkSize    = this._workers.length;
+      const chunkedPaths = [...Array(Math.ceil(input.length / chunkSize))].map(
+        (_, i) => input.slice(i * chunkSize, i * chunkSize + chunkSize));
+
+      await Promise.all(this._workers.slice().reverse().map((worker, i) => {
+        // TODO: This logic needs to be reworked. We split up the .csv files among the workers.
+        // There is a possibility a worker does not get a .csv file, therefore we need to give it an
+        // empty DataFrame.
```
mzegar, line 137: Will add an issue for this as well.

mzegar:
I think we are good to merge this, I'll put up an issue to handle the empty df a bit more pleasantly. Also another issue to separate the `createTable` methods to support more file types in the future..

PR #309, opened 2021-10-05
Add StringSeries.concatenate

bryevdv:
Pre-req for hypergraph implementation.

The Python "counterpart" for this would be `__add__` which does not exist on the JS side, so I have simply mirrored the `cudf::strings::concatenate` API from libcudf. Open to suggestion for better naming since there is already, e.g. `concat` for other things.


AjayThorve:
looks good, but I can see the naming being a bit confusing although there are enough unique factors as well.

`concat` is applied to a stringSeries, whereas `concatenate` is a standalone function accepts multiple `stringSeries` as a list? I think that should be fine

bryevdv:
looks like it did merge! 🤷

PR #323, opened 2021-10-21
Add preliminary hypergraph functions

bryevdv:
This PR adds ports of the Python `cugraph` functions for hypergraph transform.   A simple `Graph` class that does just enough to get hypergraph test running has been added. I expect we will want to expand `Graph` in the future as more use-cases come up. All major tests have been ported.

Lastly, this work differs from the Python implementation in that the hyper/direct cases are split into separate functions.


modules/cugraph/src/renumber.ts
```diff
@@ -0,0 +1,37 @@
...
+import {DataFrame, Int32, Series, StringSeries, Utf8String} from '@rapidsai/cudf';
+
+function join(edges: DataFrame, nodes: DataFrame<{id: Int32, node: Utf8String}>, col: string) {
+  const tmp = edges.assign({node: edges.get(col)}).join({other: nodes, on: ['node']}).sortValues({
+    idx: {ascending: true}
+  });
+  return edges.drop([col]).assign({[col]: tmp.get('id')});
+}
+
+export function renumber_nodes(src: StringSeries, dst: StringSeries) {
```
trxcllnt, line 24: Any particular reason these are string series? I believe src/dst could be any dtype.

bryevdv: Currently I believe there is one needed method that is only on string series (on phone atm will have to dig it up later) was going to make a separate issue top generalize

bryevdv: Well, I can't find what I was thinking of before I will give a shot at generalizing it

bryevdv: see https://github.com/rapidsai/node/pull/323/commits/e541219c493e060e5103cba79a58a2a1be3a5266

modules/cugraph/src/graph.ts
```diff
@@ -0,0 +1,53 @@
...
+import {renumber_edges, renumber_nodes} from './renumber';
+
+export interface EdgelistProps {
+  source?: string, destination?: string,
+}
+
+export class Graph {
+  df: DataFrame
+  layout: GraphCOO
+  renumber_map?: DataFrame<{id: Int32, node: Utf8String}>
```
trxcllnt, line 27: Seems coupled to the src/dst being strings still?

bryevdv: Ah for sure, none of the original tests exercise anything except the string case, so nothing ever hit a compilation issue

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+                                     nodeType);
+  const events = new DataFrame({});
+  const nodes  = entities;
+
+  const graph = create_graph(edges, source, target);
+
+  return {nodes, edges, events, entities, graph};
+}
+
+function _compute_columns(values: DataFrame, columns: string[]|null, skip: string[]) {
```
mzegar, line 201: Could columns just be `string[]` with the default being an empty array? Likewise in the interface defined above. Similar to `skip?`

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+      [name]: col,
+      [nodeId]: _prepend_str(col, cat, delim),
+      [category]: _scalar_init(cat, col.length),
+      [nodeType]: _scalar_init(name, col.length),
+    });
+    node_dfs.push(df);
+  }
+  const nodes = new DataFrame()
+                  .concat(...node_dfs)
+                  .dropDuplicates('first', true, true, [nodeId])  // correct defaults?
```
mzegar, line 264: I'm assuming still being determined?

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+                              nodeId: string,
+                              category: string,
+                              nodeType: string) {
+  const node_dfs: DataFrame[] = [];
+
+  for (const name of columns) {
+    const cat = name in categories ? categories[name] : name;
+    let col   = events.get(name);
+    col       = col.unique();
+    col       = dropNulls ? col.dropNulls() : col;  // nansToNulls?
```
mzegar, line 251: I'm assuming still being determined?

mzegar (commented):
Gave it a look over :+1: , completely outside of my knowledge though.

modules/cugraph/src/hypergraph.ts
bryevdv: Let's see what @trxcllnt says

modules/cugraph/src/hypergraph.ts
bryevdv: AFAICT these are correct, I will remove the comment 
```suggestion
                  .dropDuplicates('first', true, true, [nodeId])
```

modules/cugraph/src/hypergraph.ts
bryevdv: Just from personal experience in other contexts of having done exactly that --- used "empty list is a special value that actually means grab everything available" --- I've always regretted it, and have always ended up reverting it for the sake of consistency so that the empty list just means the empty list.

modules/cugraph/src/hypergraph.ts
bryevdv:
```suggestion
    col       = dropNulls ? col.dropNulls() : col;
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+ * ``eventId`` to a row's unique ID, ``skip`` to all non-categorical columns (or ``columns`` to all
+ * categorical columns), and ``categories`` to group columns with the same kinds of values.
+ */
+export function
+hypergraph<T extends TypeMap = any>(values: DataFrame<T>, {
+  columns       = null,
+  dropNulls     = true,
+  categories    = {},
+  dropEdgeAttrs = false,
+  skip          = [],
```
trxcllnt, line 102:
```suggestion
  skip          = [] as (keyof T & string)[],
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+ * Consider a list of events. Each row represents a distinct event, and each column some metadata
+ * about an event. If multiple events have common metadata, they will be transitively connected
+ * through those metadata values. Conversely, if an event has unique metadata, the unique metadata
+ * will turn into nodes that only have connections to the event node.  For best results, set
+ * ``eventId`` to a row's unique ID, ``skip`` to all non-categorical columns (or ``columns`` to all
+ * categorical columns), and ``categories`` to group columns with the same kinds of values.
+ */
+export function
+hypergraph<T extends TypeMap = any>(values: DataFrame<T>, {
+  columns       = null,
```
trxcllnt, line 98:
```suggestion
  columns       = values.names,
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+ *
+ * Consider a list of events. Each row represents a distinct event, and each column some metadata
+ * about an event. If multiple events have common metadata, they will be transitively connected
+ * through those metadata values. Conversely, if an event has unique metadata, the unique metadata
+ * will turn into nodes that only have connections to the event node.  For best results, set
+ * ``eventId`` to a row's unique ID, ``skip`` to all non-categorical columns (or ``columns`` to all
+ * categorical columns), and ``categories`` to group columns with the same kinds of values.
+ */
+export function hypergraphDirect<T extends TypeMap = any>(values: DataFrame<T>, {
+  columns       = null,
```
trxcllnt, line 157:
```suggestion
  columns       = values.names,
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+  categories    = {},
+  dropNulls     = true,
+  edgeShape     = {},
+  dropEdgeAttrs = false,
+  skip          = [],
```
trxcllnt, line 162:
```suggestion
  skip          = [] as (keyof T & string)[],
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+  /**
+     Dictionary mapping column names to distinct categories. If the same
+     value appears columns mapped to the same category, the transform will
+     generate one node for it, instead of one for each column.
+   */
+  categories?: {[key: string]: string};
+  /** If True, exclude each row's attributes from its edges (default: False) */
+  dropEdgeAttrs?: boolean;
+  /** A sequence of column names not to transform into nodes. */
+  skip?: string[];
```
trxcllnt, line 35:
```suggestion
  skip?: readonly (keyof T & string)[];
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+
+import {DataFrame, Int32, Series, StringSeries, Utf8String} from '@rapidsai/cudf';
+import {TypeMap} from '@rapidsai/cudf';
+
+import {GraphCOO} from './addon';
+import {renumber_edges, renumber_nodes} from './renumber';
+
+export type HypergraphBaseProps = {
+  /** An optional sequence of column names to process. */
+  columns?: string[]|null;
```
trxcllnt, line 23:
```suggestion
  columns?: readonly (keyof T & string)[];
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {DataFrame, Int32, Series, StringSeries, Utf8String} from '@rapidsai/cudf';
+import {TypeMap} from '@rapidsai/cudf';
+
+import {GraphCOO} from './addon';
+import {renumber_edges, renumber_nodes} from './renumber';
+
+export type HypergraphBaseProps = {
```
trxcllnt, line 21:
```suggestion
export type HypergraphBaseProps<T extends TypeMap = any> = {
```

modules/cugraph/src/hypergraph.ts
trxcllnt, line 35:
```suggestion
  skip?: readonly (keyof T & string)[];
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+  dropEdgeAttrs = false,
+  skip          = [],
+  delim         = '::',
+  nodeId        = 'node_id',
+  eventId       = 'event_id',
+  attribId      = 'attrib_id',
+  category      = 'category',
+  edgeType      = 'edge_type',
+  nodeType      = 'node_type',
+}: HypergraphProps = {}): HypergraphReturn {
```
trxcllnt, line 110:
```suggestion
}: HypergraphProps<T> = {}): HypergraphReturn {
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+  eventId?: string;
+  /** The name to use as the category column in the graph and DataFrames */
+  category?: string;
+  /** The name to use as the edge type column in the graph and edge DataFrame */
+  edgeType?: string;
+  /** The name to use as the node type column in the graph and node DataFrames. */
+  nodeType?: string;
+}
+
+export type HypergraphProps = HypergraphBaseProps&{
```
trxcllnt, line 50:
```suggestion
export type HypergraphProps<T extends TypeMap = any> = HypergraphBaseProps<T>&{
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+  /** The name to use as the category column in the graph and DataFrames */
+  attribId?: string;
+}
+
+export type HypergraphDirectProps = HypergraphBaseProps&{
```
trxcllnt, line 55:
```suggestion
export type HypergraphDirectProps<T extends TypeMap = any> = HypergraphBaseProps<T>&{
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+  skip          = [],
+  delim         = '::',
+  source        = 'src',
+  target        = 'dst',
+  nodeId        = 'node_id',
+  eventId       = 'event_id',
+  category      = 'category',
+  edgeType      = 'edge_type',
+  nodeType      = 'node_type',
+}: HypergraphDirectProps = {}): HypergraphReturn {
```
trxcllnt, line 171:
```suggestion
}: HypergraphDirectProps<T> = {}): HypergraphReturn {
```

modules/cugraph/src/hypergraph.ts
```diff
@@ -0,0 +1,399 @@
...
+  const nodes  = entities;
+
+  const graph = create_graph(edges, source, target);
+
+  return {nodes, edges, events, entities, graph};
+}
+
+function _compute_columns(values: DataFrame, columns: string[]|null, skip: string[]) {
+  const result: string[] = [];
+  for (const name of columns ?? values.names) {
```
trxcllnt, line 203:
```suggestion
function _compute_columns<T extends TypeMap = any>(columns: readonly(keyof T & string)[],
                                                   skip: readonly(keyof T & string)[]) {
  const result: string[] = [];
  for (const name of columns) {
```

bryevdv:
Merging, per discussion w/ @trxcllnt Issue/PR to add fixed prefix `concatenate` and categorical improvements based on that to follow separately.

PR #332, opened 2021-11-12
[FIX] set `<pkg>_ROOT` instead of `<pkg>_DIR` to work around rapids-cmake libcudacxx issue

trxcllnt:
Fixes the build for the latest versions of RAPIDS dependencies.


mzegar (approved):
Seems reasonable

PR #337, opened 2021-11-19
add Column.replaceSlice

bryevdv:
Quick stepping stone PR adds `Column.replaceSlice` that will later be used to implement categorical metadata support in CuGraph `hypergraph`


mzegar (approved):
LGTM

bryevdv:
I'll go merge this now and add categorical metadata support in a dedicated follow-on PR

PR #339, opened 2021-12-01
[WIP] Switch notebook container from nteract to jlab

bryevdv:
With this PR, running `yarn docker:run:devel:notebook` (after rebuilding) opens up the Jupyter Desktop app with an `ijskernel` installed (instead of `nteract`):

image (1)

Some issues:

* currently must `require` with absolute path, have not figured out how to get `node` to search `/opt/rapids/node/node_modules` by default
* When jlab starts, must select JS "console" &mdash; trying to start JS "notebook" opens an ipython notebook for me


bryevdv:
rebased on latest main. Still not sure why explicitly setting `NODE_PATH` does not function, or for that matter why the exact same JS kernel package is behaving any differently given the otherwise minimal/unrelated changes in the PR

bryevdv:
I had thought of that too, I've been looking but hadn't turned up anything yet. What's odd is the current reported paths:
```
require("module")._nodeModulePaths(process.cwd()).concat(require("module").globalPaths)

]
  '/opt/rapids/node_modules',
  '/opt/node_modules',
  '/node_modules',
  '/usr/local/lib/node'
]
```
That's *almost* right? 

Not sure why just the important one `/opt/rapids/node/node_modules` is missing

dev/dockerfiles/devel/main.Dockerfile
```diff
@@ -402,7 +402,7 @@ export PROMPT_COMMAND=\"history -a; \$PROMPT_COMMAND\";\n\
  && apt autoremove -y && apt clean \
  && rm -rf /tmp/* /var/tmp/*
 
-ENV NODE_PATH=/usr/local/lib/node_modules
+ENV NODE_PATH=/usr/local/lib/node_modules:/opt/rapids/node/node_modules
```
trxcllnt, line 405: Idk how docker treats `:`... maybe quoting this would work? 
```suggestion
ENV NODE_PATH="/usr/local/lib/node_modules:/opt/rapids/node/node_modules"
```

trxcllnt:
`/opt/rapids` is the homedir of the container user, and node automatically adds `$HOME/node_modules` to the require paths. I wonder if docker isn't parsing after `/usr/local/lib/node_modules` because of the colon?

dev/dockerfiles/devel/main.Dockerfile
bryevdv: This does not seem to fix. It does appear that the jlab process does not get the original shell env vars, or at least very little show up in `os.environ` in a python nb

bryevdv:
>  I wonder if docker isn't parsing after `/usr/local/lib/node_modules` because of the colon?

Well, thats the thing too, `/usr/local/lib/node`  is what is inthe global paths, ***not*** `/usr/local/lib/node_modules`  (i.e. the NODE_PATH value)

bryevdv:
Well, if I just run `bash` instead of `jlab` and then
```
rapids@bvhp:~/node$ NODE_PATH=/opt/rapids/node/node_modules/ /opt/JupyterLab/jupyterlab-desktop 
```
it *still* does not work, so it definitely seems jlab is hiding env vars or something.

bryevdv:
OK setting the working dir in the `kernel.json` *does* work :tada:

bryevdv:
OK @trxcllnt @exactlyallan  This is ready, including updated notebooks:

image (2)

PR #328, opened 2021-10-26
[DRAFT] Point-cloud streaming

mzegar:
1. Created a demo that uses default `PointCloudLayer` to do webrtc streaming 
2. Added demo to the viz app
3. Bringing in the IO module to parse the `.las` file then pass it to the deckgl context

PR #348, opened 2022-01-18
[FEA] Point-cloud streaming

trxcllnt:
Making a new PR from my fork to replace https://github.com/rapidsai/node/pull/328, because I don't have access to push git lfs objects to @mzegar's fork.


mzegar:
Sounds good, one thing I forgot to mention is I committed the DeckGL `.las` file to the repo. Might be worth getting rid of it at some point since it's 26.2 MB.
https://github.com/rapidsai/node/blob/main/modules/io/test/test.las

PR #352, opened 2022-01-20
[FIX] Update luma.gl/deck.gl versions and fix graph demo

trxcllnt:


AjayThorve (approved):
can confirm this fixes the graph demo issues

PR #362, opened 2022-02-02
Automatically disposing scope

bryevdv:
This PR adds a `scope()` function that can be used to enclose blocks of code that will return a new `Series` or `DataFrame`. All other `Series` or `DataFrame` objects created in the scope will automatically have their `dispose` methods called after. 

cc @trxcllnt tracing this code, it seems to be behaving exactly as expected/intended: a non-disposed result is returned, and all other objects in `resources` were disposed. But now tests e.g. `renumber-test.ts` are failing. Is there some subtlety of applying this that I am missing? E.g. before there was `tmp.drop(['id'])` to suppress disposal, but that is not really an option now. Could that be significant?

### DIFF
--- a/modules/cuml/src/utilities/array_utils.ts
+++ b/modules/cuml/src/utilities/array_utils.ts
@@ -16,6 +16,7 @@ import {
   DataFrame,
   Int32,
   Numeric,
+  scope,
   Series,
 } from '@rapidsai/cudf';
 
@@ -29,7 +30,7 @@ export function dataframeToSeries<T extends Numeric, K extends string>(
 }
 
 /**
- * convert a series to a dataframe as per the number of componenet in umapparams
+ * convert a series to a dataframe as per the number of components in umapparams
  * @param input
  * @param n_samples
  * @param nComponents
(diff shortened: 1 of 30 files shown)


modules/cuml/src/utilities/array_utils.ts
```diff
@@ -39,10 +40,12 @@ export function seriesToDataFrame<T extends Numeric>(
...
-  }
-  return result;
+  return scope(() => {
+    for (let i = 0; i < nComponents; i++) {
+      const indices =
+        Series.sequence({type: new Int32, init: i, size: nSamples, step: nComponents});
+      result = result.assign(new DataFrame({[i]: input.gather(indices)}));
+    }
+    return result;
+  }, [result]);
```
trxcllnt, line 50:
```suggestion
  for (let i = 0; i < nComponents; i++) {
    result = scope(() => {
      const indices =
        Series.sequence({type: new Int32, init: i, size: nSamples, step: nComponents});
      return result.assign(new DataFrame({[i]: input.gather(indices)}));
    }, [result]);
  }
  return result;
```

trxcllnt (approved):

PR #373, opened 2022-02-15
Add scope() to DataFrame.flatten

bryevdv:
Was seeing 
> cuDF failure at: /opt/rapids/node/.cache/source/cudf/0820b596e5c87006d5cc7553042c1927fe7132e3/cpp/src/column/column_view.cpp:55: Null data pointer.

until `this` was added to `doNotDispose` in the outer call to `scope`, which I suppose makes sense given the reduction involving columns of `this`, but is also slightly annoying.

### DIFF
--- a/modules/cudf/src/data_frame.ts
+++ b/modules/cudf/src/data_frame.ts
@@ -548,16 +548,16 @@ export class DataFrame<T extends TypeMap = any> {
                    const table = includeNulls ? df.asTable().explodeOuter(i, mr)  //
                                               : df.asTable().explode(i, mr);
                    return DataFrame.fromTable(table, this.names as any);
-                 }, [df]) as any;
+                 }, [this]) as any;
         },
         new DataFrame<{
           // clang-format off
-        [P in R | keyof T]:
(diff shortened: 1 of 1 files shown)


trxcllnt (changes requested):
I think we do want to dispose the intermediate DFs created by each loop of `reduce()`, but not the source or final DF.

I think something like this should work?

```typescript
    return listColumnIndices.reduce(
      (df, i, j, a) => {
        return scope(() => {
          const mr    = j === a.length - 1 ? memoryResource : undefined;
          const table = includeNulls ? df.asTable().explodeOuter(i, mr)  //
                                     : df.asTable().explode(i, mr);
          return DataFrame.fromTable(table, this.names as any);
        }, [this]) as any;
      },
      new DataFrame<{
        // clang-format off
      [P in R | keyof T]:
        P extends R
        ? T[P] extends List
          ? ListChild<T[P]>
          : T[P]
        : T[P]
        // clang-format on
      }>(this._accessor as any));

```

trxcllnt (approved):

PR #374, opened 2022-02-15
Add DedupedEdgesGraph

bryevdv:
Adds a `DedupedEdgesGraph` subclass of `Graph` that automatically removes duplicate edges. Spectral clustering algorithms that assume deduped edges are moved as well.

cc @trxcllnt once you are :+1: on the API, I'd like to flesh out the graph.ts tests more, they are currently a bit sparse in general.


trxcllnt (approved):

bryevdv:
OK, I also cleaned up and slightly expanded the graph tests (split methods into individual tests and compare actual data rather than reprs).

PR #386, opened 2022-04-21
Add readJSON.

thomcom:
This may not be required for the work I'm doing because of the jsonPath/getJSONOjbect that is cudf `String`, but here are the initial bindings, anyway!


thomcom:
Having spent some more time working with `cudf.read_json` I don't think this should be added. It supports JSON poorly enough that any end user would definitely go down a rabbit hole and not have their needs met.

PR #393, opened 2022-05-31
[FEA] Update to Arrow v8.0.0

trxcllnt:


thomcom:
Might as well figure them out on my machine, I'm likely to spend a couple more days refactoring and improving #392.

PR #402, opened 2022-06-28
Fix target/source keymapping

thomcom:
Previously, the index value referred to by a graphology graph `edges` list in the `source` and `target` values was being used to lookup `nodes` according to their position in the `nodes` array. This is incorrect.

This bugfix creates the `key=>index` mapping so that `source` and `target` refer to nodes by their `key` entry, not their list position.

This bugfix also moves the default `index.js` execution of `@rapidsai/demo-api-server` to port `3010`, compatible with the [sigma.js demo](https://www.github.com/jacomyal/sigma.js/pull/1252).

Finally this bugfix adds `.setNullMask` after the `getJSONObject` calls, as spurious entries in the nullmask are occurring there since the original merge https://github.com/rapidsai/node/pull/392

### DIFF
--- a/modules/demo/api-server/util/gpu_cache.js
+++ b/modules/demo/api-server/util/gpu_cache.js
@@ -60,7 +60,7 @@ function json_aoa_to_dataframe(str, dtypes) {
   dtypes.forEach((_, ix) => {
     const get_ix       = `[${ix}]`;
     const parse_result = tokenized.getJSONObject(get_ix);
-    parse_result.setNullMask(1, 0);
+    parse_result.setNullMask([], 0);
     arr[ix] = parse_result.cast(dtypes[ix]);
   });
   const result = new DataFrame(arr);
--- a/modules/demo/api-server/util/gpu_cache.js
+++ b/modules/demo/api-server/util/gpu_cache.js
@@ -29,10 +29,12 @@ function json_key_attributes_to_dataframe(str) {
   const no_open_list = str.split('[\n').gather([1], false);
   const tokenized    = no_open_list.split('},');
   const keys         = tokenized.getJSONObject('.key');
-  arr['key']         = keys.cast(new Int32);
+  keys.setNullMask(1, 0);
+  arr['key'] = keys.cast(new Int32);
   columns.forEach((col, ix) => {
     const parse_result = tokenized.getJSONObject('.attributes.' + columns[ix]);
-    arr[col]           = parse_result.cast(dtypes[ix]);
+    parse_result.setNullMask(1, 0);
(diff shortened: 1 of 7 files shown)


modules/demo/api-server/util/gpu_cache.js
```diff
@@ -28,10 +28,13 @@ function json_key_attributes_to_dataframe(str) {
...
   const no_open_list = str.split('[\n').gather([1], false);
   const tokenized    = no_open_list.split('},');
+  const keys         = tokenized.getJSONObject('.key');
+  keys.setNullMask(1, 0);
+  arr['key'] = keys.cast(new Int32);
   columns.forEach((col, ix) => {
     const parse_result = tokenized.getJSONObject('.attributes.' + columns[ix]);
-    const string_array = Series.new(parse_result);
-    arr[col]           = string_array.cast(dtypes[ix]);
+    parse_result.setNullMask(1, 0);
```
trxcllnt, line 36:
```suggestion
    parse_result.setNullMask([], 0);
```

modules/demo/api-server/util/gpu_cache.js
```diff
@@ -56,8 +60,8 @@ function json_aoa_to_dataframe(str, dtypes) {
   dtypes.forEach((_, ix) => {
     const get_ix       = `[${ix}]`;
     const parse_result = tokenized.getJSONObject(get_ix);
-    const string_array = Series.new(parse_result);
-    arr[ix]            = string_array.cast(dtypes[ix]);
+    parse_result.setNullMask(1, 0);
```
trxcllnt, line 63:
```suggestion
    parse_result.setNullMask([], 0);
```

AjayThorve:
@thomcom not sure if it's because of this PR or the previous one, but I cannot run the demo locally:

```bash
Error: Cannot find module '../util/schema.js'
Require stack:
- /opt/rapids/node/modules/demo/api-server/routes/root.js
```

seems like `./util/schema.js` is missing?

thomcom:
> > @thomcom not sure if it's because of this PR or the previous one, but I cannot run the demo locally:
> > ```shell
> > Error: Cannot find module '../util/schema.js'
> > Require stack:
> > - /opt/rapids/node/modules/demo/api-server/routes/root.js
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > seems like `./util/schema.js` is missing?
> 
> Oh yeah, that's extremely important, thanks for the catch.

Also @trxcllnt this makes me pretty stressed out because I see that we're not running any tests as part of CI/PR approval?

PR #424, opened 2022-08-31
`api-server` updates for 22.08

thomcom:
Getting onboarded back to `node-rapids`, I revisited `api-server` and `sigma.js` and corrected a few issues.

* Moved the hack for incorrect null counts into .cpp and removed it from .ts
* Touch up `README.md`
* Update Fastify from 3 to 4 due to dependencies.
* Correct client port in docs to 3001
* Enable misordered node/edge sets with `join` and error when they are not aligned.
* Create tests for the above 2 cases.

### DIFF
--- a/modules/cudf/src/column/json.cpp
+++ b/modules/cudf/src/column/json.cpp
@@ -21,10 +21,17 @@ namespace nv {
 
 Column::wrapper_t Column::get_json_object(std::string const& json_path,
                                           rmm::mr::device_memory_resource* mr) {
+  auto options = cudf::strings::get_json_object_options{};
+  options.set_missing_fields_as_nulls(true);
   try {
-    return Column::New(Env(),
-                       cudf::strings::get_json_object(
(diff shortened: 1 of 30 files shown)


modules/cudf/src/column/json.cpp
```diff
@@ -21,16 +21,16 @@ namespace nv {
 
 Column::wrapper_t Column::get_json_object(std::string const& json_path,
                                           rmm::mr::device_memory_resource* mr) {
+  auto options = cudf::strings::get_json_object_options{};
+  options.set_missing_fields_as_nulls(true);
```
trxcllnt, line 25: I don't think we want this? This'll set null fields to `'null'`.

trxcllnt: Or rather, we probably do want this but we want it to be an option the user can specify from JS.

thomcom:
I think this is ready to merge.

trxcllnt (approved):

PR #436, opened 2022-10-05
React/regl client app for displaying point-in-polygon point clouds.

thomcom:
This app will display 2d point clouds that are returned from an `api-server` I/O routine.

The app will display a background image that scales and translates with the GL view, displaying points that are contained within the viewport and culling the rest.

Part of the demo will provide an option to disable view translation but retain the viewport culling, so that points are culled from the center of the display, demonstrating the culling.

Steps remaining:
- [ ] Add mouseEvents for translation and scale
- [ ] Bind translation and scale to GL view
- [ ] Bind translation and scale to background image
- [ ] Add a large datasource to `api-server`
- [ ] Add an `api-server` endpoint to feed data via arrow
- [ ] Load points buffer in app via arrow
- [ ] Introduce some kind of legend
- [ ] Introduce an options control panel opposite to legend.


modules/demo/particles/yarn.lock
```diff
@@ -0,0 +1,4903 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
trxcllnt, line 1: should remove this file before merging

yarn.lock
```diff
@@ -7241,9 +7241,15 @@ fast-json-stringify@^2.5.2:
     string-similarity "^4.0.1"
 
 fast-json-stringify@^5.0.0:
+<<<<<<< HEAD
```
trxcllnt, line 7244: bad merge?

thomcom: Yes, already cleaned it up.

PR #473, opened 2025-10-18
# [FEA] Support ARM64 & update to RAPIDS 25.02, CUDA 12.8, Ubuntu 24.04 - Phase 1

aucahuasi:
Modernizes node-rapids to RAPIDS 25.02, CUDA 12.8, and Ubuntu 24.04 with ARM64 (aarch64) support for GH200 Grace Hopper platforms.

## Changes

* Update to RAPIDS 25.02, CUDA 12.8, Ubuntu 24.04, Python 3.12
* Add ARM64 (aarch64) support alongside x86_64
* Update Arrow 9.0.0 to 19.0.0 (enable S3, Acero)
* Update nvcomp 2.4.1 to 4.2.0.11 with ARM64 binaries
* Update build system: cmake-js 7.3.1, node-gyp 10.2.0, CMake 3.30.5
* Update TypeScript 4.5.5 to 5.3.3, Jest 26.5.3 to 29.7.0
* Update @typescript-eslint 5.30.0 to 6.21.0 for TypeScript 5.3 compatibility
* Update RMM bindings for RAPIDS 25.02 API changes (`thrust::optional` to `std::optional`, removed deprecated methods)
* Remove BlazingSQL module (abandoned upstream)

## Testing

Phase 1 modules (`core`, `cuda`, `rmm`) all passing tests on Ubuntu 24.04 (x86_64 and ARM64), CUDA 12.8, Python 3.12, Node.js 16.15.1.

**Phase 2** (separate PR) will address `cudf` module with its significant RAPIDS 25.02 API changes. Node.js version was kept at 16.x for this phase; Phase 2 may target Node.js 20.x depending on testing and compatibility requirements.


modules/sql/.vscode/launch.json
trxcllnt, line 1: Does this update remove the blazingsql bindings? That makes me sad 😢

modules/core/bin/install-deps/debian.sh
```diff
@@ -101,7 +101,7 @@ if [ -n "$APT_DEPS" ]; then
     sudo apt update
     sudo apt install -y $APT_DEPS;
     if [ -n "$INSTALLED_CLANGD" ]; then
-        sudo update-alternatives --install /usr/bin/clangd clangd /usr/bin/clangd-17 100
+        sudo update-alternatives --install /usr/bin/clangd clangd /usr/bin/clangd-12 100
```
trxcllnt: clangd-12 is incredibly old and likely broken for modern CUDA. clang-format-12 will also most likely regress from clang-format-17, which is not ideal. Is there a reason we're downgrading these?

aucahuasi: Good catch, this was part of my old tests, I updated it to `clang 18` (which is compatible with `ubuntu 24.04`)

modules/sql/.vscode/launch.json
aucahuasi: I share your sadness about removing BlazingSQL. As you know, I was the first technical engineer and tech lead at BlazingSQL for years, and we collaborated on some great work together back in the early RAPIDS/cuDF days.

Since it's unmaintained upstream, I wasn't sure whether to try porting it or remove it. This PR is already complex with `ARM64` + `RAPIDS 25.02` + `CUDA 12.8` + `Ubuntu 24.04`.

What's the RAPIDS team's policy: should unmaintained modules stay but be disabled, or is removal the right approach?

trxcllnt: Yeah I understand why you removed it. We'd been [using a fork](https://github.com/rapidsai/node/blob/968e09ba16fafee650d79213708a7154d37d873d/modules/core/cmake/Modules/ConfigureBlazingSQL.cmake#L46) I maintained and updated as necessary to match changes in libcudf. It wasn't difficult, but I imagine it would be a large amount of work to get it up to this libcudf version.

aucahuasi:
I've successfully updated the toolchain (`LLVM 18`, `GCC 13`, `sccache 0.10.0`, `Node.js 16.20.2`) and tested the `core`, `cuda`, and `rmm` modules in the devel container. Everything builds and runs correctly with `CUDA 12.8`.

A couple of questions:

1. **Version bumping**: Should external contributors handle version updates across the monorepo, or do you prefer to manage this during your release process?
2. **Packages workflow**: The `yarn docker:build:devel:packages` command fails trying to access `S3` credentials. Is this workflow meant for RAPIDS internal use only, or should external contributors be able to run it? If the latter, the `package.Dockerfile` would need to support local-only `sccache` (currently `sccache 0.10.0` tries AWS autodiscovery even with empty `S3` config).

For now, I've verified the modules build and work in the devel container. Let me know what else you need for review!

trxcllnt:
@aucahuasi thanks for this PR, I'll try to review it this weekend. Typically we'd update the RAPIDS version across all the projects, but I should be able to handle that. There's also some housekeeping tasks, like sccache, that I can tackle.

I'll push any updates to this branch if you don't mind, then approve and merge once I've double checked everything still works on my end.

aucahuasi:
> @aucahuasi thanks for this PR, I'll try to review it this weekend. Typically we'd update the RAPIDS version across all the projects, but I should be able to handle that. There's also some housekeeping tasks, like sccache, that I can tackle.
> 
> I'll push any updates to this branch if you don't mind, then approve and merge once I've double checked everything still works on my end.

Thanks @trxcllnt! That's perfect! Please go ahead and push any updates to the branch. I appreciate you handling the version updates and `sccache` configuration.

For context on the `Node.js 16.20.2` choice: it aligns with our production environment requirements for `ARM64/GH200` integration work.

Let me know if you need any additional work/testing or info from my end!

package.json
```diff
@@ -130,41 +129,8 @@
...
-    "**/@math.gl/core": "3.6.3",
-    "**/apache-arrow": "^9.0.0",
-    "**/gl-matrix": "3.4.3",
-    "**/jsdom": "16.6.0",
-    "**/math.gl": "3.6.3",
-    "**/mjolnir.js": "2.7.1",
-    "**/maplibre-gl": "2.4.0"
+    "@types/babel__traverse": "~7.14.0",
+    "**/@types/minimatch": "5.1.2"
   },
```
trxcllnt, line 134: @aucahuasi was there a reason to remove these pins from the workspace? We had these because it's very easy for npm to install incompatible variations of these modules that break when different versions are loaded at runtime, so we needed to enforce the same versions in all node_modules.

aucahuasi: No reason, I think I could not build at that moment and that made the trick! I don't see any reason not to keep those if we want it, I can re-create the dev env to test again and revert that change.

trxcllnt:
Closing in favor of https://github.com/rapidsai/node/pull/474 (which includes the commits in this PR). Thanks again @aucahuasi!

